PR #28, opened 2014-01-30
Make mock available by default

sgonyea:
- The require will not trigger until redis_mock accessed.
- If we happen to call `SidekiqUniqueJobs.redis_mock.flushdb`
  in a `before(:each)` but `testing_enabled?` hasn't yet been
  called, tests will all blow up. This avoids the need for an
  extra `require` in the test/spec helper.
- Switches to ruby's built-in `defined?` operator.

### DIFF
--- a/spec/lib/sidekiq_testing_enabled_spec.rb
+++ b/spec/lib/sidekiq_testing_enabled_spec.rb
@@ -100,6 +100,27 @@ describe "Sidekiq::Testing" do
   end # describe 'Enabling/Disabling Mock Redis in Tests'
 
   describe "When Sidekiq::Testing is enabled" do
+    describe 'After unique jobs have been performed' do
+      before { SidekiqUniqueJobs::Config.unique_args_enabled = true }
+      after  { SidekiqUniqueJobs::Config.unique_args_enabled = false }
+
+      let(:payload) { 'foobar' }
+      let(:payload_hash) { SidekiqUniqueJobs::PayloadHelper.get_payload("UniqueWorker", "working", [payload]) }
+
+      # This test is failing
+      it "should remove the unique lock for the job" do
+        expect {
+          UniqueWorker.perform_async(payload)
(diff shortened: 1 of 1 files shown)


lib/sidekiq-unique-jobs/middleware/client/unique_jobs.rb
```diff
@@ -71,12 +71,7 @@ def payload_hash
         # When sidekiq/testing is loaded, the Sidekiq::Testing constant is
         # present and testing is enabled.
         def testing_enabled?
-          if Sidekiq.const_defined?('Testing') && Sidekiq::Testing.enabled?
-            require 'sidekiq-unique-jobs/testing'
-            return true
-          end
-
-          false
+          SidekiqUniqueJobs.testing_enabled? && SidekiqUniqueJobs.use_redis_mock?
```
mhenrixon, line 74: I see you already did :) Sorry was following old stale code.

lib/sidekiq-unique-jobs/testing.rb
```diff
@@ -1,6 +1,29 @@
-require 'mock_redis'
 module SidekiqUniqueJobs
   def self.redis_mock
-     @redis_mock ||= MockRedis.new
+     @redis_mock ||= begin
+       require 'mock_redis'
+       MockRedis.new
```
midu, line 5: Any reason why you wouldn't be using the same redis instance that sidekiq uses instead? If the users want to use mockredis, they can configure sidekiq to use mockredis. 

This has the expected (to me) effect of storing sidekiq-unique-jobs' keys in another instance of mockredis. I clear my mock redis before each test, but the unique keys stay in this other instance of mockredis between my tests...

I know that it was already an issue before your PR, but you seem to have a good grasp on the problems of testing with sidekiq unique jobs and maybe the change would make sense to you.

mhenrixon, line 5: Good point!

PR #75, opened 2015-01-27
Checking Sidekiq::Testing.inline? on testing strategy and connector

### LINKED ISSUE #71: Sidekiq::Testing inline detection assumes you're always using inline testing
The check in `lib/sidekiq_unique_jobs/config.rb` only looks if you're using sidekiq testing, not which method you're using. So, even if you're using sidekiq's fake testing method, sidekiq-unique-jobs still blows with the mock_redis gem. Is this gem necessary in all scenarios?

Draiken:
When checking the testing strategy's `eligible?` method the `Sidekiq::Testing.inline?` was being checked, but not on the testing connector.

This caused mock_redis to be required even when inline testing was not being used.

Should fix #71

### DIFF
--- a/lib/sidekiq_unique_jobs/config.rb
+++ b/lib/sidekiq_unique_jobs/config.rb
@@ -30,8 +30,8 @@ module SidekiqUniqueJobs
       end
     end
 
-    def testing_enabled?
-      if Sidekiq.const_defined?('Testing') && Sidekiq::Testing.enabled?
+    def inline_testing_enabled?
+      if Sidekiq.const_defined?('Testing') && Sidekiq::Testing.enabled? && Sidekiq::Testing.inline?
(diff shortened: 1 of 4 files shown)


Draiken:
Just a note, since unique jobs uses redis to lock/unlock jobs and gets the connection through sidekiq, the tests involving classes with `unique: true` will actually try to use redis.

Not sure if this should be in readme or something.

I believe to completely remove this issue, unique jobs should implement a strategy that doesn't actually use redis when inline testing is being used (the same way sidekiq does it).

felixyz:
I needed this patch to get things working without having to modify my test suite – a point release would be great so I don't have to use master! @mhenrixon 

(All in all, is it really worth the trouble using mock_redis? Especially since Sidekiq itself has other strategies for isolating tests.)

Issue #77, opened 2015-03-31
Jobs are unlocked if they fail and are retried


DarthMax:
I just discovered that jobs are unlocked if they fail. This happens regardless if there are retries left or if the job dies. I'm wondering if this is the intended behavior. As I would understand (and need) that fact a job that will be retired should still be unique.

The code responsible for this is the following:  *(in  /lib/sidekiq_unique_jobs/middleware/server/unique_jobs.rb)*

```ruby
def call(worker, item, _queue, redis_pool = nil)
  ...
  yield
  ensure
    if after_yield? || !defined? unlocked || unlocked != 1
      unlock(lock_key)
    end
end
```

So are there any reasons for this behavior or do I miss anything?

__Update__

So after some research I start to understand why unlocking works the way it does. I was not aware of the fact that jobs in the schedule/retry queue are pushed to the worker queues via client push (involving sidekiqs client middlewares).  If in that situation the job is still locked the job will never be pushed into its worker queue. 

To circumvent this issue it might be possible to save the jid of the job that acquired the lock instead of the [1,2]. So the client middleware can check for the jid and let the job reenqueue if they match.

What do you think of this idea?

mhenrixon:
That sounds like a great idea for fallback when the job had been originally failed or scheduled.

mhenrixon:
So a few updates concerning this went live with `v3.0.13`. Would you mind trying that and see if the problem still persists? If you are still having a problem just give me a failing test and I will fix it.

mhenrixon:
This was closed by #96 and a few recent commits.

PR #87, opened 2015-06-15
Do not unlock on sidekiq shutdown

deltaroe:
somewhat related to https://github.com/mhenrixon/sidekiq-unique-jobs/issues/77

The current behavior of sidekiq-unique-jobs is to unlock a job when job is done regardless of if the job succeeds or fails.  From what I can tell this is desired because sidekiq requeues failed jobs and sidekiq-unique-jobs would otherwise prevent them from being queued.

There is one exception and that is when a task is killed due to exceeding a timeout during a worker shutdown. [https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/processor.rb#L55-L59](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/processor.rb#L55-L59)  In that case the job is left and the worker picks it back up when it restarts.  The task kill generates an exception and sidekiq-unique-job removes the lock even though the job is still there. When the worker starts again it picks up the job again, but non-unique work can be queued because the lock was already removed. 

This PR duplicates the logic of the sidekiq processor to also catch the shutdown exception and not unlock the job in that case.

I first noticed this when having a unique reoccurring task that took longer than the reoccurrence (ran for 100 seconds, but started every minute) If the worker was restarted while the task was running it would start up again and at the top of the next minute a duplicate task would start. When the first task would complete it would unlock the lock placed by the second task and the two jobs would leapfrog forever.

### DIFF
--- a/lib/sidekiq_unique_jobs/middleware/server/unique_jobs.rb
+++ b/lib/sidekiq_unique_jobs/middleware/server/unique_jobs.rb
@@ -8,6 +8,7 @@ module SidekiqUniqueJobs
         attr_reader :unlock_order, :redis_pool
 
         def call(worker, item, _queue, redis_pool = nil)
+          shutdown = false
           @redis_pool = redis_pool
 
           decide_unlock_order(worker.class)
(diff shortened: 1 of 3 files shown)


mhenrixon:
Looks great! Thanks for this

Issue #72, opened 2015-01-16
Runtime uniqueness when using :before_yield as unlock order


tsubery:
I have a specific use case for sidekiq uniquness for a project I work on. We need the possibility to enqueue more jobs while a job is running but we never want two unique jobs to run concurrently. As far as i understand there is no way to do it as :after_yield unlock order would not allow to enqueue more jobs while a worker is running and :before_yield would allow simultaneous workers. I made a fork adding a locking mechanism for each running job to fix our problem. You can find it here: 
https://github.com/tsubery/sidekiq-unique-jobs/tree/runtime_uniqueness
I also added some documentation describing it. 
Do you think that is something that might be interesting to other users of the gem?

mhenrixon:
This might fix the problem some people have with uniqueness and help get rid of some of the confusion I think.

mhenrixon:
@tsubery ? Any pull request coming for this feature? :)

tsubery:
The way I implemented it was using another gem called redis-lock.
Before starting a job i would just acquire lock and release it after the job was done or timeout occured.
I don't recall all the specific details but I remember encountering problems with some of the dependencies and ended up patching both redis-lock and mock-redis to make the tests pass.
It is very likely I'm missing some knowledge to create an appropriate solution but it makes more sense to me to use a lock inside the specific job than to use my implementation that relies on patching two extra gems.
I would be very curious to hear about alternative solutions.

stevenjonescgm:
I had a similar use case (wanted the job to be unique in the queue but also able to reschedule itself). I solved/mitigated by writing another worker that simply queued the first so the uniqueness could bounce between them.

Before I noticed ```unique_unlock_order: :before_yield```, I tried to be clever with my FutureWorker and set ```unique_job_expiration: 0.seconds``` so that FutureWorker could be rescheduled as soon as ```item[at]``` was passed. This caused an exception to be raised from setex: "ERR invalid expire time in setex" because when the scheduled job was picked up, it gets moved to the active queue and attempts to set expire time < 1.

I'll open a pr, but I'm no longer certain that it is a useful change since ```unique_unlock_order: :before_yield``` is better for my case and Sidekiq::Worker.perform_in protects from ```item['at'] < Time.now```

pik:
The current `before_yield` is completely arbitrary. Basically it will have a near unique behaviour if the jobs are fast to run and the queue is long and absolutely no-uniqueness guarantees in the opposite scenario.

[...]

PR #92, opened 2015-07-03
Add after unlock hook

HParker:
Found myself with a similar problem to @stevenjonescgm. I need runtime uniqueness, but the ability for a worker to requeue itself. My solution was to add an `after_unlock` hook where you could requeue your work after the lock was released.

Should help with https://github.com/mhenrixon/sidekiq-unique-jobs/issues/72


deltaroe:
Remerged with upstream master

Issue #94, opened 2015-07-26
Jobs can unlock mutexes they don't own 


pik:
This happens with perform_at or when a job runs longer than the key_expire time, this is also explicitly advised against in the [redis docs](http://redis.io/topics/distlock).

mhenrixon:
Am I understanding it correctly that the new job gets unlocked by a newer job is there something I am missing?

pik:
So at the moment jobs do not check if they were the ones which originally set a mutex when unlocking -- meaning there are conditions where the job will delete a mutex that another job has set (if it's own has already expired).

mhenrixon:
Closed by #96

PR #96, opened 2015-07-27
Prevent Jobs from deleting mutexes they don't own

pik:
Hello - so this should solve #94. The jid should already be sufficiently unique for each job and both server and client middleware have access to this. 

Using the LUA script would be preferable since it saves a back-and-forth trip to the server, but if there users of sidekiq-unique-jobs with old versions of redis -- perhaps it would make sense to use the conn.watch version by default. (It's included for the tests anyways since mock-redis does not support lua scripts).

### DIFF
--- a/lib/sidekiq_unique_jobs/lib.rb
+++ b/lib/sidekiq_unique_jobs/lib.rb
@@ -1,0 +1,23 @@
+require 'pry-byebug'
+module SidekiqUniqueServerLib
+  def remove_if_matches
+    <<-LUA
+      if redis.call('GET', KEYS[1]) == ARGV[1] then
+        redis.call('DEL', KEYS[1])
+      end
+    LUA
+  end
+end
+
+module SidekiqUniqueServerMockLib
+  def unlock(lock_key, item)
+    connection do |con|
+      con.watch(lock_key)
+      if con.get(lock_key) == item['jid']
+        ret = con.multi { con.del(lock_key) }
+      else
+        con.unwatch
+      end
+    end
+  end
+end
--- a/lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb
+++ b/lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb
@@ -66,7 +66,7 @@ module SidekiqUniqueJobs
               else
                 unique = conn.multi do
                   # set value of 2 for scheduled jobs, 1 for queued jobs.
-                  conn.setex(payload_hash, expires_at, item['at'] ? 2 : 1)
+                  conn.setex(payload_hash, expires_at, item['jid'])
                 end
               end
             end
@@ -75,7 +75,7 @@ module SidekiqUniqueJobs
 
           def new_unique_for?
             connection do |conn|
-              return conn.set(payload_hash, item['at'] ? 2 : 1, nx: true, ex: expires_at)
+              return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at)
             end
           end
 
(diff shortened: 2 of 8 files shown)


lib/sidekiq_unique_jobs/lib.rb
```diff
@@ -0,0 +1,23 @@
+require 'pry-byebug'
```
mhenrixon, line 1: Please remove this :)

lib/sidekiq_unique_jobs/middleware/server/unique_jobs.rb
```diff
@@ -48,12 +50,12 @@ def after_yield?
 
         protected
 
-        def payload_hash(item)
-          SidekiqUniqueJobs::PayloadHelper.get_payload(item['class'], item['queue'], item['args'])
+        def unlock(lock_key, item)
+          connection {|c| c.eval(remove_if_matches, :keys => [lock_key], :argv => [item['jid']]) }
```
mhenrixon, line 54: In unique jobs we use the newer hash style `keys: [lock_key]` etc.

lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb
```diff
@@ -75,7 +75,7 @@ def old_unique_for?
 
           def new_unique_for?
             connection do |conn|
-              return conn.set(payload_hash, item['at'] ? 2 : 1, nx: true, ex: expires_at)
+              return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at)
```
mhenrixon, line 78: How is jid used? The original functionality ensured that we can use both scheduled (timed) jobs and the regular immediate queue. Won't this be a problem now in regards to uniqueness? I thought the 2 and 1 is used by sidekiq to determine if it is a scheduled type of job or a regular? Am I missing something @pik ?

lib/sidekiq_unique_jobs/lib.rb
```diff
@@ -0,0 +1,23 @@
...
+module SidekiqUniqueServerLib
+  def remove_if_matches
+    <<-LUA
```
mhenrixon, line 4: Great idea using LUA for this!

lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb
pik, line 78: @mhenrixon Am I missing something? 
https://github.com/pik/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb#L78
Is not identical to: https://github.com/pik/sidekiq-unique-jobs/blob/bug/unlock-mutex/lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb#L62
(Since SETNX will always fail if any value is set.) Was the idea to let simultaneous scheduling and queuing of jobs? I was under the impression this was deprecated but lacking code-comments it's hard to tell the intent? 

>  I thought the 2 and 1 is used by sidekiq to determine if it is a scheduled type of job or a regular?

As far as I understand Sidekiq just uses time['at'] to schedule into the future:   https://github.com/mperham/sidekiq/blob/master/lib%2Fsidekiq%2Fclient.rb#L186 
Where could the value Sidekiq-Unique is storing into a custom hash be accessible to Sidekiq?  

The purpose of the patch storing the `jid` as the key value is to keep Jobs from unlocking a mutex which it did not lock (if it's own key was cleared or expired).

mhenrixon, line 78: > (Since SETNX will always fail if any value is set.) Was the idea to let simultaneous scheduling and queuing of jobs? I was under the impression this was deprecated but lacking code-comments it's hard to tell the intent? 

Those are equivalent since it is being used with watch or at least that was the intention.

PR #99, opened 2015-08-05
Add run_lock to prevent duplicate running jobs

pik:
#72 This patch adds a `:run_lock` along side `:before_yield` and `:after_yield` unlock ordering. `:run_lock` should guarantee that only a single unique Job is ever running at one time - without preventing another job from being queued. The default run_lock options are as follows: 
```
      default_run_lock_retry_interval: 0,
      default_run_lock_retries: 0,
      default_reschedule_on_lock_fail: false,
      default_run_lock_expire: 60,
```
if reschedule_on_lock_fail is set the job will be pushed back into the queue (perhaps reschedule_in could also be an option to push it into the scheduled queue?). If run_lock_retries is set it will spin_lock for run_lock_retries after initially failing to acquire the lock with run_lock_retry_interval as the wait time between attempts (@tsubery). 

On failure to acquire a lock it will Raise a RunLockFailedError. I'm not sure if this is preferable to failing silently since if the job has Sidekiq `:retries` options set it will effectively reschedule anyways; but I do think that reschedule on lock fail should be done by SidekiqUniqueJobs rather than through bubbling an error up to the Sidekiq processor since the later will consume a retry attempt, which might not be intended.

### DIFF
--- a/lib/sidekiq_unique_jobs/exception.rb
+++ b/lib/sidekiq_unique_jobs/exception.rb
@@ -1,0 +1,3 @@
+module SidekiqUniqueJobs
+  class RunLockFailedError < StandardError; end
+end
--- a/lib/sidekiq_unique_jobs/server/middleware.rb
+++ b/lib/sidekiq_unique_jobs/server/middleware.rb
@@ -7,45 +7,82 @@ module SidekiqUniqueJobs
     class Middleware
       include Extensions
 
-      attr_reader :unlock_order, :redis_pool
+      attr_reader :unlock_order,
+                  :redis_pool,
+                  :reschedule_on_lock_fail,
+                  :run_lock_retries,
+                  :run_lock_retry_interval
 
-      def call(worker, item, _queue, redis_pool = nil)
-        operative = true
+      def call(worker, item, _queue, redis_pool = nil, &blk)
         @redis_pool = redis_pool
-        decide_unlock_order(worker.class)
+        setup_options(worker.class)
+        send("#{unlock_order}_call", item, &blk)
+      end
+
+      def run_lock_call(item)
         lock_key = payload_hash(item)
-        unlock(lock_key, item) if before_yield?
+        run_lock = try_acquire_run_lock(lock_key, item)
+        if run_lock
+          unlock(lock_key, item)
+          yield
+        else
+          if reschedule_on_lock_fail
+            reschedule(item)
+          else # Not sure if we want to raise?
+            raise SidekiqUniqueJobs::RunLockFailedError
+          end
+        end
+      ensure
+        if run_lock
+          unlock_run(lock_key, item)
+        end
+      end
+
+      def before_yield_call(item)
+        unlock(payload_hash(item), item)
+        yield
+      end
+
+      def after_yield_call(item)
+        operative = true
         yield
       rescue Sidekiq::Shutdown
         operative = false
         raise
       ensure
-        unlock(lock_key, item) if after_yield? && operative
+        unlock(payload_hash(item), item) if operative
       end
 
-      def decide_unlock_order(klass)
-        @unlock_order = if unlock_order_configured?(klass)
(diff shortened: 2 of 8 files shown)


lib/sidekiq_unique_jobs/server/middleware.rb
```diff
@@ -7,45 +7,86 @@ module Server
...
       end
 
-      def default_unlock_order
-        SidekiqUniqueJobs.config.default_unlock_order
+      def setup_options(klass)
+        @options = klass.get_sidekiq_options if klass.respond_to?(:get_sidekiq_options)
+      end
+
+      def unlock_order
+        if !options[:unique]
```
mhenrixon, line 69: I prefer early returns like.

``` ruby
def unlock_order
  return :never unless options[:unique]
end
```

lib/sidekiq_unique_jobs/server/middleware.rb
```diff
@@ -67,6 +108,33 @@ def logger
...
+
+      # rubocop:disable HandleExceptions
+      def try_acquire_run_lock(lock_key, item)
+        status = begin
+          (run_lock_retries + 1).times do
+            status = connection do |con|
+              con.set("#{lock_key}:run", item['jid'], nx: true, ex: run_lock_expire)
+            end
+            break if status
+            sleep(run_lock_retry_interval)
```
mhenrixon, line 130: Sleeping sounds like a terrible idea, it could potentially take down the whole infrastructure of someones system with some simple misconfiguration in some place and would be very hard to track why. Is this really required?

lib/sidekiq_unique_jobs/server/middleware.rb
```diff
@@ -7,45 +7,86 @@ module Server
...
+        setup_options(worker.class)
+        send("#{unlock_order}_call", item, &blk)
+      end
+
+      # rubocop:disable MethodLength
+      def run_lock_call(item)
         lock_key = payload_hash(item)
-        unlock(lock_key, item) if before_yield?
+        run_lock = try_acquire_run_lock(lock_key, item)
+        if run_lock
```
mhenrixon, line 26: I prefer if we could get rid of all the nested ifs. Lets try and do early returns to avoid nesting.

lib/sidekiq_unique_jobs/exception.rb
```diff
@@ -0,0 +1,3 @@
+module SidekiqUniqueJobs
+  class RunLockFailedError < StandardError; end
```
mhenrixon, line 2: can we rename this file to `run_lock_failed_error.rb` instead? I prefer it if a class or module has its own file. Potentially we could hide them under an `exceptions` or `errors` module if they get in the way of navigating the rest of the files.

lib/sidekiq_unique_jobs/server/middleware.rb
mhenrixon, line 130: A better approach to me would be to push this back for later processing but I haven't delved into the details of the code I just feel that would be a safer approach. It is also a feature that people have asked for being able to just delay a locked job until the currently running one is done. `Reschedulable` jobs I suppose it would be called.

pik, line 130: Yeah, looking at it again I don't think the spin-lock is very nice, we could use a blocking attempt to get a mutex with a timeout (either one of these will idle a Sidekiq Thread for the interval though).

Maybe simply reschedule or drop if the lock fails on first try?

mhenrixon:
Not bad at all I think we are onto something here. @pik would love to get together on Skype if you have some time. I can be found as `mhenrixon`.

lib/sidekiq_unique_jobs/server/middleware.rb
mhenrixon, line 130: I think reschedulable jobs would be better. I'll do some work on that if you could see about the code style and whatever other comments I had!

Issue #98, opened 2015-08-03
Confusing behavior when trying to `[1,2,3].each { |n| SomeJob.perform_in(n.seconds.from_now, n) }` never running, logging as duplicate value


just3ws:
I'm in a Rails 4.2 project running Sidekiq 3.4.2, Redis-rb 3.2.1, Hiredis 0.6.0, Sidekiq-Unique-Jobs 3.0.14 and am running Redis server version 3.0.3. 

Step to repo:

- Stop Sidekiq
- Stop Console
- Execute `flushall` in redis-cli (and verify that database has been cleaned)
- Start Sidekiq (and see that Sidekiq has set up it's state in Redis)
- Have job that looks like
```
class MainJob
  include Sidekiq::Worker
  sidekiq_options unique: true, log_duplicate_payload: true
  def perform(status_id)
    Rails.logger.debug("RUN #{status_id} - #{Time.zone.now}")
  end
end
```
- Run in console ` [ 1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20].each { |x|  MainJob.perform_in(x.seconds.from_now, x);}`
- See the logs report everything as being duplicate
```
22:03:03 sidekiq.1 | started with pid 52254
22:03:05 sidekiq.1 | 2015-08-03T03:03:05.658Z 52254 TID-ow0dcxq00 INFO: Booting Sidekiq 3.4.2 with redis options {:url=>"redis://127.0.0.1:6379", :namespace=>"sidekiq_twitterstats_development"}
22:03:05 sidekiq.1 | 2015-08-03T03:03:05.967Z 52254 TID-ow0dcxq00 INFO: Running in ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
22:03:05 sidekiq.1 | 2015-08-03T03:03:05.967Z 52254 TID-ow0dcxq00 INFO: See LICENSE and the LGPL-3.0 for licensing details.
22:03:05 sidekiq.1 | 2015-08-03T03:03:05.967Z 52254 TID-ow0dcxq00 INFO: Upgrade to Sidekiq Pro for more features and support: http://sidekiq.org/pro
22:03:05 sidekiq.1 | 2015-08-03T03:03:05.973Z 52254 TID-ow0dcxq00 INFO: Starting processing, hit Ctrl-C to stop
[...]

Issue #93, opened 2015-07-23
Jobs scheduled in the future are never run


amiel:
Here's what I think is happening:

1. When the job is scheduled (`Worker.perform_in 1.minute, args`), it runs through the [clients client middleware](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/middleware.rb#L17), setting the unique lock.
2. Once the scheduled time is up, it runs through the [servers client middleware](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/middleware.rb#L10) and considered not unique.

The [servers client middleware](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/middleware.rb#L8-11) was introduced in ea06539afd0f126bd934b8018cfa557b2f6cf095. Could it be safely removed? Or am I misunderstanding the issue?

amiel:
BTW, I'm happy to work on a pull-request for this, but wanted to start a discussion first to make sure I understand the context.

mhenrixon:
Think @pik is onto something in regards to this problem. Keep an eye out for #96

amiel:
@mhenrixon thanks for the update.

mhenrixon:
Could you post a failing test for this @amiel ?

amiel:
@mhenrixon I'll give it a shot. Do you have a suggestion as to where this test should live?

mhenrixon:
Just add it to the issue description and tell me when its done. Oh and please base it on current master!

amiel:
@mhenrixon ok, will do.

amiel:
@mhenrixon I'm having a hard time replicating this issue in tests.

However, I can replicate it with the current master.

Here's what I've done to troubleshoot:

1. I have a worker with the following options:
    
   ```ruby
   sidekiq_options queue: :low, retry: 2, unique: true, log_duplicate_payload: true,
   unique_args: :unique_args, unique_job_expiration: 1.day, unique_unlock_order: :never
   ```

   Note: my `unique_args` just uses the first argument (it ignores the second argument).

2. In addition to enabling `log_duplicate_payload`, I have added a debugging statement in `MyWorker#perform`.

3. Everything seems to work as expected using `perform_async`:

   ```ruby
   MyWorker.perform_async 1
   # => jid
   # Server log shows `MyWorker#perform` was called
[...]

Issue #91, opened 2015-07-02
perform_at and perform_async do not unique if perform_at is in the future.


HParker:
When you queue the same jobs twice, once using `perform_async` and once using `perform_at` with the same arguments. They are not counted as duplicates if the time provided to `perform_at` is in the future.

Example:
```ruby
pry(main)> MyUniqueWorker.perform_at(Time.now + 1.minute, 1)
=> "81f090cda6659b529ff91e5a"
[6] pry(main)> MyUniqueWorker.perform_async(1)
=> "d1469ddf0c6f9fada9875db2"
```

There are now **two** jobs that have the same arguments and should have been de-duped.

My unique worker is equivalent to:
```ruby
class MyUniqueWorker
  include Sidekiq::Worker
  sidekiq_options queue: :default,
                  retry: true,
                  unique: true,
                  unique_job_expiration: 2.hours,
                  retry_count: 10

  def perform(id)
    # do some work...
  end
end
```

## Work Around

The work around I am using now is to override `perform_in` and `perform_async` so that the 'at' arg is always there.
```ruby
# override perform_async to just schedule for now.
  # due to unique jobs bug.
  def self.perform_async(*args)
    perform_in(Time.now, *args)
  end

  # override perform_in to never remove the 'at' key
  def self.perform_in(interval, *args)
    int = interval.to_f
    now = Time.now
    ts = (int < 1_000_000_000 ? (now + interval).to_f : int)

    # removed the optimization in lib/sidekiq/worker.rb
    # item.delete('at'.freeze) if ts <= now.to_f
    item = { 'class' => self, 'args' => args, 'at' => ts }
    client_push(item)
  end
```

## Long Term Solution

Any ideas at a long term solution for this?
Looking through the code it looks like something around [*_unique_for?](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/middleware/client/strategies/unique.rb#L57) might be causing this issue.
Any ideas for a fix would be appreciated.

HParker:
Found a PR that seems to relate: https://github.com/mhenrixon/sidekiq-unique-jobs/pull/85

mhenrixon:
Since the below tests pass fine using real redis I am closing this issue.

```ruby
class MyUniqueWorker
  include Sidekiq::Worker
  sidekiq_options queue: :customqueue, retry: true, unique: true,
                  unique_job_expiration: 7200, retry_count: 10
[...]

PR #101, opened 2015-08-09
Fix scheduled jobs breaking with new_unique_for method

pik:
#93 #98 So having dug into the code it looks like `set_nx` was not a viable option for `new_unique_for` and neither was my update to the old `unique_for` whereby it was altered to always use item['jid'] without a special assignment if it's being scheduled rather than enqueued. This patch fixes the broken behaviour for both `new_unique_for` and `old_unique_for`.

### DIFF
--- a/lib/sidekiq_unique_jobs/client/extensions.rb
+++ b/lib/sidekiq_unique_jobs/client/extensions.rb
@@ -4,7 +4,7 @@ module SidekiqUniqueJobs
       def lock_queue_script
         <<-LUA
           local ret = redis.call('GET', KEYS[1])
-          if not ret or ret == 'scheduled' then
+          if not ret or string.sub(ret, 1, 9) == 'scheduled' then
             return redis.call('SETEX', KEYS[1], ARGV[1], ARGV[2])
           end
         LUA
--- a/lib/sidekiq_unique_jobs/client/extensions.rb
+++ b/lib/sidekiq_unique_jobs/client/extensions.rb
@@ -1,0 +1,14 @@
+module SidekiqUniqueJobs
+  module Client
+    module Extensions
+      def lock_queue_script
+        <<-LUA
+          local ret = redis.call('GET', KEYS[1])
+          if not ret or ret == 'scheduled' then
+            return redis.call('SETEX', KEYS[1], ARGV[1], ARGV[2])
+          end
+        LUA
+      end
+    end
+  end
+end
--- a/lib/sidekiq_unique_jobs/client/middleware.rb
+++ b/lib/sidekiq_unique_jobs/client/middleware.rb
@@ -1,9 +1,12 @@
-require 'sidekiq_unique_jobs/server/middleware'
 require 'sidekiq_unique_jobs/connectors'
+require 'sidekiq_unique_jobs/client/extensions'
+require 'sidekiq_unique_jobs/server/middleware'
 
 module SidekiqUniqueJobs
   module Client
     class Middleware
+      include Extensions
+
       def call(worker_class, item, queue, redis_pool = nil)
         @worker_class = SidekiqUniqueJobs.worker_class_constantize(worker_class)
         @item = item
@@ -44,14 +47,14 @@ module SidekiqUniqueJobs
       def old_unique_for?
         connection do |conn|
           conn.watch(payload_hash)
-          pid = conn.get(payload_hash).to_i
-          if pid == 1 || (pid == 2 && item['at'])
+          pid = conn.get(payload_hash)
+          if pid && (pid != 'scheduled' || item['at'])
             conn.unwatch
             nil
           else
             conn.multi do
               if expires_at > 0
-                conn.setex(payload_hash, expires_at, item['jid'])
+                conn.setex(payload_hash, expires_at, item['at'] ? 'scheduled' : item['jid'])
               else
                 conn.del(payload_hash)
               end
@@ -62,8 +65,18 @@ module SidekiqUniqueJobs
       # rubocop:enable MethodLength
 
       def new_unique_for?
+        item['at'] ? unique_for_schedule : unique_for_queue
+      end
+
+      def unique_for_schedule
         connection do |conn|
-          return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at)
+          conn.set(payload_hash, 'scheduled', nx: true, ex: expires_at)
+        end
+      end
+
+      def unique_for_queue
+        connection do |conn|
+          conn.eval(lock_queue_script, keys: [payload_hash], argv: [expires_at, item['jid']])
         end
       end
 
--- a/lib/sidekiq_unique_jobs/client/middleware.rb
+++ b/lib/sidekiq_unique_jobs/client/middleware.rb
@@ -42,38 +42,48 @@ module SidekiqUniqueJobs
         SidekiqUniqueJobs.config.unique_storage_method
       end
 
-      # rubocop:disable MethodLength
       def old_unique_for?
+        item['at'] ? unique_schedule_old : unique_enqueue_old
+      end
+
+      def unique_enqueue_old
         connection do |conn|
           conn.watch(payload_hash)
-          pid = conn.get(payload_hash)
-          if pid && (pid != 'scheduled' || item['at'])
-            conn.unwatch
-            nil
-          else
+          lock_val = conn.get(payload_hash)
+          if !lock_val || lock_val[0..8] == 'scheduled'.freeze
             conn.multi do
-              if expires_at > 0
-                conn.setex(payload_hash, expires_at, item['at'] ? 'scheduled' : item['jid'])
-              else
-                conn.del(payload_hash)
-              end
+              conn.setex(payload_hash, expires_at, item['jid'])
             end
+          else
+            conn.unwatch
+            false
           end
         end
       end
-      # rubocop:enable MethodLength
 
-      def new_unique_for?
-        item['at'] ? unique_for_schedule : unique_for_queue
-      end
-
-      def unique_for_schedule
+      def unique_schedule_old
         connection do |conn|
-          conn.set(payload_hash, 'scheduled', nx: true, ex: expires_at)
+          conn.watch(payload_hash)
+          if expires_at < 0 || conn.get(payload_hash)
+            conn.unwatch
+            false
+          else
+            conn.setex(payload_hash, expires_at, "scheduled-#{item['jid']}")
+          end
         end
       end
 
-      def unique_for_queue
+      def new_unique_for?
+        item['at'] ? unique_schedule : unique_enqueue
+      end
+
+      def unique_schedule
+        connection do |conn|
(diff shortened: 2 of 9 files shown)


lib/sidekiq_unique_jobs/client/middleware.rb
```diff
@@ -44,14 +46,14 @@ def storage_method
       def old_unique_for?
         connection do |conn|
           conn.watch(payload_hash)
-          pid = conn.get(payload_hash).to_i
-          if pid == 1 || (pid == 2 && item['at'])
+          pid = conn.get(payload_hash)
+          if pid && (pid != 'scheduled' || item['at'])
```
mhenrixon: I'm sorry but `if != ||` == WAT?!

lib/sidekiq_unique_jobs/sidekiq_unique_ext.rb
```diff
@@ -25,7 +25,8 @@ def payload_hash(item)
 
       def unlock(lock_key, item)
         Sidekiq.redis do |con|
-          con.eval(remove_on_match, keys: [lock_key], argv: [item['jid']])
+          val = @parent && @parent.name == 'schedule' ? 'scheduled' : item['jid']
```
mhenrixon: Please create a method for this value and name it appropriately

lib/sidekiq_unique_jobs/client/middleware.rb
```diff
@@ -44,14 +46,14 @@ def storage_method
...
             conn.unwatch
             nil
           else
             conn.multi do
               if expires_at > 0
-                conn.setex(payload_hash, expires_at, item['jid'])
+                conn.setex(payload_hash, expires_at, item['at'] ? 'scheduled' : item['jid'])
```
mhenrixon: Why are we not using the `jid` if the job is scheduled? Like `scheduled:jid` and then check if the key starts with scheduled?

lib/sidekiq_unique_jobs/client/extensions.rb
```diff
@@ -0,0 +1,14 @@
+module SidekiqUniqueJobs
+  module Client
+    module Extensions
+      def lock_queue_script
+        <<-LUA
+          local ret = redis.call('GET', KEYS[1])
+          if not ret or ret == 'scheduled' then
```
mhenrixon: This would have to change as well then with a composite key of some sort. Even leaving it as it was with just a simple integer. Not sure what happens in an application with a billion scheduled keys. The jit was needed to prevent unlocking jobs you shouldn't be able to unlock?

lib/sidekiq_unique_jobs/client/middleware.rb
mhenrixon: Code was a mess already. What I'd like to see is something more in the lines of:

``` ruby
def old_unique_for?
  connection do |conn|
    conn.watch(payload_hash)
    pid = conn.get(payload_hash)
    return create_lock(conn) unless pid == 'scheduled' || item['at']
    conn.unwatch
  end
end

def create_lock(conn)
  conn.multi do
    return clear_exired_lock(conn) unless expires_at > 0
    conn.setex(payload_hash, expires_at, item['at'] ? 'scheduled' : item['jid'])
  end
end

def clear_expired_lock(conn)
  conn.del(payload_hash) # already expired
end
```

Thoughts on that?

lib/sidekiq_unique_jobs/client/middleware.rb
mhenrixon: Or is the jid not at all interesting until it ends up in the real queue? I have been traveling for 60 hours and feel a little sluggish

mhenrixon:
Love the contribution! Had a few notes but if something is out of order or doesn't make sense just say the word and I will spend some time when I am back from the dead.

pik:
@mhenrixon Thanks for the feedback on this. I'm going to try and go through my understanding of what's happening just to make sure this is the correct fix and it's not introducing any subtle bugs. 

The [old_unique_for](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/686b7d7235/lib%2Fsidekiq_unique_jobs%2Fmiddleware%2Fclient%2Fstrategies%2Funique.rb#L53) never removed the locks for a scheduled job - since it was possible to just promote (over-write) a scheduled lock to a queued-lock (I think these were `1` and `2` respectively). (The alternative would be to have a Mutex for schedule and a Mutex for queue, but this seems more involved as I'm not sure there is an easy way to tell when a Job is being queued whether it had previously been scheduled or not? (once `item['at']` has been deleted).

As far as I understand the Sidekiq behaviour:
Job has item['at'] sent - Sidekiq client `atomic_push` sends it to the `schedule` queue if `item['at'] != nil`, middleware is called once. At schedule time when the job is fetched if `item['at']` is in the past this is deleted and middleware is called again.

So the `old_unique_for` behaviour would have been (pseudo-code):
```
Middleware run 1 (client):
  # Mutex is nil
If not Mutex or Mutex is 1 ('scheduled') and not item['at']:
  Set Mutex to -> 1 ('scheduled')
  Added to 'schedule' queue
else:
  # Job is not scheduled

Middleware run 2 (client):
# Mutex is 1
# Item['at'] has been deleted
If not Mutex or Mutex is 1 ('scheduled') and not item['at']:
  Set Mutex to -> 2 (normal queue)
  Added to the get_sidekiq_options['queue'] queue
else:
  # Do not Schedule another job, or add a duplicate job to a normal queue.

Middleware run 3 (server):
  # Jobs are always trying to Unlock, even if Unique: True is not set.
  # Issue #100
  # Not checking whether Job Timeout and Mutex belongs to it or not #Fixed by checking JID
  If before_yield:
    Set Mutex -> Nil
    # If other jobs are queued and fetched after this they can start to run at the same-time with this one
    Run Job
  Else:
    Run Job
    # If Jobs are dropped while this is running some values might remain out-dated
    Set Mutex -> Nil
``` 

The reason SETNX was not working with the `new_unique_for` is because SETNX is always false if any value is present. Meaning that it would have looked
like this for step 2 (above):
```
Middleware run 2 (client):
# Mutex is 1
# Item['at'] has been deleted
If SETNX(<key>, 2):
  # This fails
  # Payload is considered duplicate
  Set Mutex to -> 2 (normal queue)
  Added to the get_sidekiq_options['queue'] queue
else:
  # Do not Schedule another job, or add a duplicate job to a normal queue.
```

New Version:

```
Middleware run 1 (client):
# Mutex is nil
If not Mutex or Mutex is 'scheduled'(1) and not item['at']:
  Set Mutex to -> 'scheduled' (1)
  Added to 'schedule' queue
else:
  # Job is not scheduled


Middleware run 2 (client):
# Mutex is 'scheduled' (1)
# Item['at'] has been deleted
If not Mutex or Mutex is 'scheduled' (1) and not item['at']:
  Set Mutex to -> item ['jid']
  Added to the get_sidekiq_options['queue'] queue
else:
  # Do not Schedule another job, or add a duplicate job to a normal queue.
```

  # Not checking whether Job Timeout and Mutex belongs to it or not,
  # Fixed by checking JID b51e733cbde1c8d8d844e819f307e194bf042e18
```
Middleware run 3 (server):
  If before_yield:
    # Compare JID
    If Mutex Set By Me:
      Set Mutex -> Nil
    # If other jobs are queued and fetched after this they can start to run at the same-time with this one
    Run Job
  else if after_yield:
    Run Job
    # If Jobs are dropped while this is running some values might remain out-dated
    # Compare JID
    If Mutex Set By Me:
      Set Mutex -> Nil
```
RunLock PR: https://github.com/mhenrixon/sidekiq-unique-jobs/pull/99
```
  else if run_lock:
    # Optional spin-lock to acquire
    If acquire Run-Mutex:
      Set Run-Mutex -> Item['jid']
      Set Mutex -> Nil
      Run Job
      # Another job can still be queued, but will not run till this is complete
      Set Run-Mutex -> Nil
    Else:
      Drop or Reschedule
  else:
    # Not using Unique
    # Run Job
```
https://github.com/pik/sidekiq-unique-jobs/commit/74bafc42b05656ebd29e3bae50d5d94033b82d74#diff-40bba6fd31628d2a95fd0ab341dcde73R42

mhenrixon:
There is still a problem with #98. The thing we need to do is to create a composite key with the time to run the job and the argument. We need to then make sure that the jobs unique arguments are used for locking the run as to prevent the same job running twice (simultaneously) if the first job gets delayed.

Does that make sense at all? I don't see the code taking this into consideration.

pik:
@mhenrixon Are you sure? All of the jobs in #98 run correctly for me with this patch. 

> We need to then make sure that the jobs unique arguments are used for locking the run as to prevent the same job running twice (simultaneously) if the first job gets delayed.

This should still happen because an JID type lock can only over-ride a 'scheduled' lock, so when the 'scheduled' job comes up and runs middleware again either it can promote it's previous 'scheduled' lock to an JID or a JID will already be there and it will fail. I can try and add some more specs for this?

lib/sidekiq_unique_jobs/client/middleware.rb
pik: So we wouldn't normally need the `jid` in the schedule - because SidekiqUnique never unlocks `scheduled` only promotes to the enqueued lock -- the exception being deletion of jobs from the scheduled queue via. the sidekiq API after for example a Sidekiq restart that would have allowed the lock to expire (thus would then allow a scheduled job to delete the `scheduled` mutex of another job). 

Over all I'm not sure if it's worth caring that much about uniqueness in the scheduled queue, so I may submit a patch later to not have any kind of scheduled uniqueness by default -- but probably we should be as consistent as possible if it is going to be supported? 

Another to thing to note is using something like "scheduled-#{item['jid']}" will require a substring call in the LUA script - which is more expensive than a direct comparison.

deltaroe:
I'd like to propose another solution.  https://github.com/mhenrixon/sidekiq-unique-jobs/pull/105

pik:
@deltaroe That doesn't solve anything - please see failing specs, I added a reversion for it in #107. 

Also this branch is not passing only on versions of Sidekiq older than 3.1 - I believe this has to do with a bug in the redis connection pool implementation for watch/multi/exec on those versions of Sidekiq. Since this problem doesn't exist on the newer versions of Sidekiq I think it would just make sense to set the minimum requirement to 3.2 (Sidekiq is at 3.4.2 now and a lot of things have changed since).

deltaroe:
@pix It sounds like we have different use cases. In our environment we don't want a job to be able to run now if it's scheduled to run in the future. This was brought up by my colleague in https://github.com/mhenrixon/sidekiq-unique-jobs/issues/91, however PR https://github.com/mhenrixon/sidekiq-unique-jobs/pull/96 changed the behavior to what we expected and was merged between when the issue was reported and @mhenrixon tried to reproduce the issue. 

I wonder if adding some method to specify which mode is desired would make sense here

pik:
@deltaroe #96 only changed the  unique for to work the same way as the `new_unique_for` method - I was unaware at the time that the `new_unique_for` method was responsible for the scheduling issue -- hence this current PR.

PR #105, opened 2015-08-19
Allow job with jid matching unique lock to pass unique check

deltaroe:
Related to https://github.com/mhenrixon/sidekiq-unique-jobs/pull/101

### DIFF
--- a/lib/sidekiq_unique_jobs/client/middleware.rb
+++ b/lib/sidekiq_unique_jobs/client/middleware.rb
@@ -63,7 +63,7 @@ module SidekiqUniqueJobs
 
       def new_unique_for?
         connection do |conn|
-          return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at)
+          return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at) || conn.get(payload_hash) == item['jid']
         end
       end
 
--- a/spec/lib/client/middleware_spec.rb
+++ b/spec/lib/client/middleware_spec.rb
@@ -71,10 +71,10 @@ describe SidekiqUniqueJobs::Client::Middleware do
 
     it 'enqueues previously scheduled job' do
       QueueWorker.sidekiq_options unique: true
-      QueueWorker.perform_in(60 * 60, [1, 2])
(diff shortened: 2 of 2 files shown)


spec/lib/client/middleware_spec.rb
```diff
@@ -71,10 +71,10 @@ def perform(_)
 
     it 'enqueues previously scheduled job' do
       QueueWorker.sidekiq_options unique: true
-      QueueWorker.perform_in(60 * 60, [1, 2])
+      jid = QueueWorker.perform_in(60 * 60, 1, 2)
```
deltaroe, line 74: This test wasn't testing the same job moving from the scheduled to its normal queue. The signature for the perform_in method is `def perform_in(interval, *args)` https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/worker.rb#L45  Therefore the first job had a single argument of an array and the second job had two Fixnum arguments. 

Removing the two brackets caused the test to fail identifying issue #101

lib/sidekiq_unique_jobs/client/middleware.rb
```diff
@@ -63,7 +63,7 @@ def old_unique_for?
 
       def new_unique_for?
         connection do |conn|
-          return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at)
+          return conn.set(payload_hash, item['jid'], nx: true, ex: expires_at) || conn.get(payload_hash) == item['jid']
```
pik, line 66: I don't think this works - because then a perform_async job cannot run if a perform_in job is scheduled (even if the scheduled job is hours in advance). I'd rather not have a lock on the scheduled queue at all then a behaviour like this.

deltaroe, line 66: @pik It sounds like we have different use cases. In our environment we don't want a duplicate job to be able to run now if it's scheduled in the future even if it's several hours out.

pik, line 66: @deltaroe This is how Sidekiq-unique-jobs always handled jobs before -- it's not a "fix" if you implement a strategy that works for you and breaks it for everyone else.

PR #107, opened 2015-08-21
Revert broken fix and add failing specs

pik:
Revert for #105 which did not fix anything.
Add failing specs to master then it would be preferable for everyone to know that the current Master is broken. 

(If this is not promptly being resolved #101 has been outstanding for a while now)  (Currently our application is on 3.0.13, which still has unresolved problems but not as bad as the current Head).

### DIFF
--- a/spec/lib/client/middleware_spec.rb
+++ b/spec/lib/client/middleware_spec.rb
@@ -71,10 +71,10 @@ describe SidekiqUniqueJobs::Client::Middleware do
 
     it 'enqueues previously scheduled job' do
       QueueWorker.sidekiq_options unique: true
-      jid = QueueWorker.perform_in(60 * 60, 1, 2)
+      QueueWorker.perform_in(60 * 60, [1, 2])
 
       # time passes and the job is pulled off the schedule:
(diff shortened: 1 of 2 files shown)


spec/lib/client/middleware_spec.rb
```diff
@@ -71,10 +95,10 @@ def perform(_)
 
     it 'enqueues previously scheduled job' do
       QueueWorker.sidekiq_options unique: true
-      jid = QueueWorker.perform_in(60 * 60, 1, 2)
+      QueueWorker.perform_in(60 * 60, [1, 2])
```
deltaroe: I think this test is still broken and isn't testing for a job moving from the scheduled to its normal queue. 

Right now this line is creating a scheduled job with a single argument of a two element array and the later `Sidekiq::Client.push` is creating a job with two Fixnum arguments. It's not the same job and the test is only showing that two separate jobs can be created, not that the same job can move from the scheduled to a normal queue. 

Changing this line to `QueueWorker.perform_in(60 * 60, 1, 2)` should be enough to demonstrate the issue

pik: Yeah that should be 1,2 not `[1,2]`.

PR #116, opened 2015-10-06
Clean up version 4 upgrade instructions

albertyw:
I thought that the upgrade instructions would look nicer with a few things moved around.


mhenrixon:
Thanks! Looked much better in my markdown editor :)

PR #143, opened 2015-10-31
Calculate worker's unique args when a proc or a symbol is specified

### LINKED ISSUE #138: Jobs not getting queued in v4
This DOES NOT work `4.0.7`

```ruby
# All these calls are fired within a minute
TestWorker.perform_in(1.minute, 23, "skskks") # This gets queued
TestWorker.perform_in(1.minute, 23, "skskks") # This does not get queued - Expected
TestWorker.perform_in(1.minute, 3633, "skskks") # This DOES NOT get queued??? Broken?
```

```ruby
class TestWorker < BaseWorker
  sidekiq_options queue: :test,
                  unique: :until_executed,
                  unique_args: :unique_args

  def self.unique_args(id, blob)
    [id]
  end

  def perform(id, blob)
    puts "TestWorker - #{id}"
  end
end
```

This works `3.0.15`

```ruby
# All these calls are fired within a minute
TestWorker.perform_in(1.minute, 23, "skskks") # This gets queued
TestWorker.perform_in(1.minute, 23, "skskks") # This does not get queued - Expected
TestWorker.perform_in(1.minute, 3633, "skskks") # This gets queued
```

```ruby
class TestWorker < BaseWorker
  sidekiq_options queue: :test,
                  unique: true,
                  unique_args: :unique_args

  def self.unique_args(id, blob)
    [id]
  end

  def perform(id, blob)
    puts "TestWorker - #{id}"
  end
end
```

zeqfreed:
Hi!

There was a bug which made it so job's unique digest was calculated incorrectly when worker's unique_args option was specified. If unique_args was any thuthy value (e.g. a proc or a symbol), then at line 23 in file lib/sidekiq_unique_jobs/unique_args.rb nothing happened and unique_args calculation actually never took place.

I fixed this line and modified specs to test unique_digest generation for different cases of worker options.
Also I had to fix two existing test worker classes, because they were written with assumption that arguments are passed separately, but in fact in current implementation unique_args method or proc receive arguments as a single array, which by the way contradicts some examples in README file.

This PR should fix #138.

### DIFF
--- a/spec/jobs/custom_queue_job_with_filter_method.rb
+++ b/spec/jobs/custom_queue_job_with_filter_method.rb
@@ -1,7 +1,7 @@
 class CustomQueueJobWithFilterMethod < CustomQueueJob
   sidekiq_options unique: :until_executed, unique_args: :args_filter
 
-  def self.args_filter(*args)
+  def self.args_filter(args)
     args.first
   end
 end
--- a/spec/lib/sidekiq_unique_jobs/unique_args_spec.rb
+++ b/spec/lib/sidekiq_unique_jobs/unique_args_spec.rb
@@ -4,6 +4,44 @@ RSpec.describe SidekiqUniqueJobs::UniqueArgs do
   let(:item) { { 'class' => 'UntilExecutedJob', 'queue' => 'myqueue', 'args' => [[1, 2]] } }
   subject { described_class.new(item) }
 
+  context '#unique_digest' do
+    let(:item) { item_options.merge('args' => [1, 2, 'type' => 'it'] ) }
+
+    shared_context 'unique digest' do
+      context 'given another item' do
+        let(:another_subject) { described_class.new(another_item) }
+
+        context 'with the same unique args' do
+          let(:another_item) { item_options.merge('args' => [1, 2, 'type' => 'it']) }
+          it 'equals to unique_digest for that item' do
+            expect(subject.unique_digest).to eq(another_subject.unique_digest)
+          end
+        end
+
+        context 'with different unique args' do
+          let(:another_item) { item_options.merge('args' => [1, 3, 'type' => 'that']) }
+          it 'differs from unique_digest for that item' do
+            expect(subject.unique_digest).to_not eq(another_subject.unique_digest)
+          end
+        end
+      end
+    end
+
+    context 'when unique_args is a proc' do
+      let(:item_options) { {'class' => 'UntilExecutedJob', 'queue' => 'myqueue',
+                            'unique_args' => Proc.new { |args| args[1] }} }
+
+      include_context 'unique digest'
+    end
+
+    context 'when unique_args is a symbol' do
+      let(:item_options) { {'class' => 'UniqueJobWithFilterMethod', 'queue' => 'myqueue',
+                            'unique_args' => :filtered_args} }
+
+      include_context 'unique digest'
+    end
+  end
+
   describe '#unique_args_enabled?' do
     with_default_worker_options(unique: :until_executed, unique_args: ->(args) { args[1]['test'] }) do
       with_sidekiq_options_for(UntilExecutedJob, unique_args: :unique_args) do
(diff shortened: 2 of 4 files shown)


spec/jobs/custom_queue_job_with_filter_method.rb
```diff
@@ -1,7 +1,7 @@
 class CustomQueueJobWithFilterMethod < CustomQueueJob
   sidekiq_options unique: :until_executed, unique_args: :args_filter
 
-  def self.args_filter(*args)
+  def self.args_filter(args)
```
mhenrixon, line 4: I obviously forgot to remove splatting of arguments in a few places!

spec/lib/sidekiq_unique_jobs/unique_args_spec.rb
```diff
@@ -4,6 +4,44 @@
   let(:item) { { 'class' => 'UntilExecutedJob', 'queue' => 'myqueue', 'args' => [[1, 2]] } }
   subject { described_class.new(item) }
 
+  context '#unique_digest' do
+    let(:item) { item_options.merge('args' => [1, 2, 'type' => 'it'] ) }
+
+    shared_context 'unique digest' do
```
mhenrixon: This should be `shared_examples` no?

spec/lib/sidekiq_unique_jobs/unique_args_spec.rb
```diff
@@ -4,6 +4,44 @@
...
+          end
+        end
+      end
+    end
+
+    context 'when unique_args is a proc' do
+      let(:item_options) { {'class' => 'UntilExecutedJob', 'queue' => 'myqueue',
+                            'unique_args' => Proc.new { |args| args[1] }} }
+
+      include_context 'unique digest'
```
mhenrixon: And this should preferably say `it_behaves_like 'unique digest'`? It isn't just a shared context it has the actual tests so let change this up a bit. See the suggestion in the next comment.

spec/lib/sidekiq_unique_jobs/unique_args_spec.rb
```diff
@@ -4,6 +4,44 @@
   let(:item) { { 'class' => 'UntilExecutedJob', 'queue' => 'myqueue', 'args' => [[1, 2]] } }
   subject { described_class.new(item) }
 
+  context '#unique_digest' do
```
mhenrixon: I find it more intention revealing to use `describe '#unique_digest'` for when speccing methods. The whole things should preferably look more like:

``` ruby
describe '#unique_digest' do
    let(:item) { item_options.merge('args' => [1, 2, 'type' => 'it'] ) }

    shared_examples 'unique digest' do
      context 'given another item' do
        let(:another_subject) { described_class.new(another_item) }

        context 'with the same unique args' do
          let(:another_item) { item_options.merge('args' => [1, 2, 'type' => 'it']) }
          it 'equals to unique_digest for that item' do
            expect(subject.unique_digest).to eq(another_subject.unique_digest)
          end
        end

        context 'with different unique args' do
          let(:another_item) { item_options.merge('args' => [1, 3, 'type' => 'that']) }
          it 'differs from unique_digest for that item' do
            expect(subject.unique_digest).to_not eq(another_subject.unique_digest)
          end
        end
      end
    end

    context 'when unique_args is a proc' do
      let(:item_options) do
        { 'class' => 'UntilExecutedJob', 'queue' => 'myqueue',
          'unique_args' => Proc.new { |args| args[1] }}
      end
      it_behaves_like 'unique digest'
    end

    context 'when unique_args is a symbol' do
      let(:item_options) do
        { 'class' => 'UniqueJobWithFilterMethod', 'queue' => 'myqueue',
          'unique_args' => :filtered_args }
      end
      it_behaves_like 'unique digest'
    end
  end
```

mhenrixon:
Thank you for the contribution, was a great find!

PR #158, opened 2016-01-26
Fix test overrides.

benseligman:
Tests using this code failed with an undefined method error on the call to `unlock`. Since all we need from `Unlockable` is `unlock` it seems simpler and clearer to just explicitly alias the method.


bronson:
Hm, the Rubocop failures are unrelated to this change.  Master has been broken for a month: https://travis-ci.org/mhenrixon/sidekiq-unique-jobs

PR #164, opened 2016-02-15
Run lock timeout configurable

Slania:
```
class ConfigurableRunLockTimeoutWorker
  include Sidekiq::Worker
  sidekiq_options :queue => 'background', :run_lock_expiration => 60 * 3
end
```


Slania:
We use sidekiq jobs to do things like send out emails through slow APIs. With batching on our end, their end, XML processing etc., it is actually very common for some of our workers to take a couple of minutes or more to send out a large batch (~4000) emails.

Right now, the run_lock is hardcoded to expire in 60seconds unless the lease is renewed. This PR makes it possible to configure the expiration time.

Slania:
@mhenrixon thoughts?

rails_example/app/workers/simple_worker.rb
```diff
@@ -3,7 +3,7 @@ class SimpleWorker
   sidekiq_options unique: :until_executed,
                   unique_args: -> (args) { [args.first] },
                   queue: :default,
-                  default_expiration: 3 * 60 * 60
+                  default_queue_lock_expiration: 3 * 60 * 60
```
mhenrixon, line 6: Still thinking about this but maybe `default_queue_expiration` and `default_run_expiration` would suffice. We already know it is expiration of a lock so it feels redundant. On the other hand it is very explicit. What do you think?

redis/synchronize.lua
```diff
@@ -1,13 +1,14 @@
 local unique_key     = KEYS[1]
 local time           = ARGV[1]
+local expires        = ARGV[2]
```
mhenrixon, line 3: nice, I did something like this originally.

spec/jobs/long_running_job.rb
```diff
@@ -0,0 +1,7 @@
+class LongRunningJob
+  include Sidekiq::Worker
+  sidekiq_options queue: :customqueue, retry: true, unique: :until_and_while_executing,
+                  run_lock_expiration: 7_200, retry_count: 10
```
mhenrixon, line 4: Ok got it with these names it makes for sense writing out the full thing with `default_yada`.

lib/sidekiq_unique_jobs/timeout_calculator.rb
```diff
@@ -1,34 +1,21 @@
 module SidekiqUniqueJobs
-  class TimeoutCalculator
```
mhenrixon, line 2: I would honestly prefer inheritance over composition in this case because you could clean up the child calculators a bit.  Do you see it?

lib/sidekiq_unique_jobs/queue_lock_timeout_calculator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def initialize(item)
+      @item = item
+    end
+
+    def seconds
+      time_until_scheduled + queue_lock_expiration
+    end
+
+    def queue_lock_expiration
```
mhenrixon: could we make these two methods the same so that it quacks like a `TimeoutCalculator` that responds to a `calculator.lock_expiration` instead? Feels overly specific with the run_lock_expiration method when the class already conveys that and we don't really care much about  this. Maybe even just putting it in seconds if that is what is used in the end. Feels like we have a few methods too many in these classes.

``` ruby
module SidekiqUniqueJobs
  class TimeoutCalculator
    def self.for_item(item)
      new(item)
    end

    def initialize(item)
      @item = item
    end

    def time_until_scheduled
      scheduled = item[AT_KEY]
      return 0 unless scheduled
      (Time.at(scheduled) - Time.now.utc).to_i
    end

    def seconds
      raise NotImplementedError
    end

    def worker_class_queue_lock_expiration
      worker_class_expiration_for QUEUE_LOCK_TIMEOUT_KEY
    end

    def worker_class_run_lock_expiration
      worker_class_expiration_for RUN_LOCK_TIMEOUT_KEY
    end

    def worker_class
      @worker_class ||= SidekiqUniqueJobs.worker_class_constantize(item[CLASS_KEY])
    end

    private

    def worker_class_expiration_for(key)
      return unless worker_class.respond_to?(:get_sidekiq_options)
      worker_class.get_sidekiq_options[key]
    end

    attr_reader :item
  end
end

module SidekiqUniqueJobs
  class RunLockTimeoutCalculator < TimeoutCalculator
    def seconds
      @seconds ||= (
        worker_class_run_lock_expiration ||
        SidekiqUniqueJobs.config.default_run_lock_expiration
      ).to_i
    end
  end
end

module SidekiqUniqueJobs
  class QueueLockTimeoutCalculator  < TimeoutCalculator
    def seconds
      queue_lock_expiration + time_until_scheduled
    end

    def queue_lock_expiration
      @queue_lock_expiration ||=
        (
          worker_class_queue_lock_expiration ||
          SidekiqUniqueJobs.config.default_queue_lock_expiration
        ).to_i
    end
  end
end
```

If I am not completely off with all the flu, fever, coughing and sneezing that should do pretty much the same thing as your code but is a lot less to maintain :)

mhenrixon:
Good job, had some feedback for you @Slania. Good thinking with the naming of the lock_expiration and the default_lock_expiration. :+1:

lib/sidekiq_unique_jobs/queue_lock_timeout_calculator.rb
Slania: Seems fair to me :)

Slania:
@mhenrixon I think I have it. I used the same file for all 3 calculators just because it made sense to keep them close together and the file itself is relatively small. However, they each have their own specs though. Inheritance did make the testing a lot easier and more terse! :)

Issue #140, opened 2015-10-22
Broken compatibility with Sidekiq 3.4


morganatwishpond:
https://github.com/mhenrixon/sidekiq-unique-jobs/commit/cd1954248474c2c31f35c4fab4a2e8820b3b0a8b#diff-de9f106e9f00c47071520a6db18167cc breaks API compatibility with Sidekiq 3.4.1 when it comes to deleting jobs.

Specifically, https://github.com/mperham/sidekiq/blob/v3.4.1/lib/sidekiq/api.rb#L372 requires two params while https://github.com/mhenrixon/sidekiq-unique-jobs/commit/cd1954248474c2c31f35c4fab4a2e8820b3b0a8b#diff-de9f106e9f00c47071520a6db18167ccR71 accepts no params.

```
ArgumentError: wrong number of arguments (2 for 0)
	from gems/sidekiq-unique-jobs-4.0.7/lib/sidekiq_unique_jobs/sidekiq_unique_ext.rb:72:in `delete_ext'
	from gems/sidekiq-3.4.1/lib/sidekiq/api.rb:372:in `delete'
	from gems/sidekiq-unique-jobs-4.0.7/lib/sidekiq_unique_jobs/sidekiq_unique_ext.rb:23:in `delete_ext'
```

mhenrixon:
Thanks for reporting, I'll have a look at it!

warmwind:
I got exactly the same error

cnluzhang:
Does this problem still appear in Sidekiq 4.x?

Thanks.

zedtux:
+1 on this issue.

zedtux:
Using `sidekiq` 4.0.2 and `sidekiq-cron` 4.2 all is working fine.

theprogrammerin:
@mhenrixon I think #167 that should fix the problem. Can you please check.

mhenrixon:
Looks like this should be fixed now!

PR #167, opened 2016-02-17
Fix for sidekiq delete failing for version 3.4.x 

theprogrammerin:
Changelog: 

Sidekiq API till 3.4.2 https://github.com/mperham/sidekiq/blob/v3.4.2/lib/sidekiq/api.rb#L372 passes to arguments for delete. which was throwing Argument error till now. 

Fix for #140


mhenrixon:
Cheers!

PR #174, opened 2016-03-18
Troubleshooting

mathieujobin:
I'm having issues with job failing to re-acquire the lock
they are defined like this.

```ruby
class LovelyWorker
  include Sidekiq::Worker
  sidekiq_options retry: false, backtrace: true, unique: :until_executing

  def perform(object_uuid)
    do_lovely_stuff
    LovelyWorker.perform_in(15.seconds, object_uuid)
  end
end
```

I've split the return codes of the lua scripts, and added some debugging to understand a bit better what is going on... I was wondering if this added information could be useful to others, so I am opening a pull request.

The error looks like this

```
sidekiq.log:444461:2016-03-18T00:17:55.525Z 20899 TID-10xnnk SeasonMaintainerWorker JID-515066574f2af0660cc440bb DEBUG: failed to acquire lock for uniquejobs:17042957acafcc6506ef21a15f371999 (stored_jid did not match job_id) -- {"class"=>"PreGameWorker", "args"=>["2b9028c3-c482-4300-89fe-64e133aad607"], "at"=>1458261900.0, "retry"=>false, "queue"=>"default", "backtrace"=>true, "unique"=>:until_executing, "jid"=>"f0e5d10cffbb6a1d56e03cdd", "created_at"=>1458260275.5235777, "unique_prefix"=>"uniquejobs", "unique_args"=>["2b9028c3-c482-4300-89fe-64e133aad607"], "unique_digest"=>"uniquejobs:17042957acafcc6506ef21a15f371999"}
```

What I am noticing from this log line..
the class name of the job trying to get the lock is not the same one as the one who apparently owns the lock.

and these workers don't even have similar arguments, so it should have never match...

I'm not too sure what is going on, any help will be appreciated

thanks

### DIFF
--- a/redis/acquire_lock.lua
+++ b/redis/acquire_lock.lua
@@ -7,6 +7,7 @@ if stored_jid then
   if stored_jid == job_id then
     return 1
   else
+    -- maybe we should do something special in this case?
     return 0
   end
 end
--- a/redis/acquire_lock.lua
+++ b/redis/acquire_lock.lua
@@ -13,7 +13,7 @@ end
 
 if redis.pcall('set', unique_key, job_id, 'nx', 'ex', expires) then
   redis.pcall('hsetnx', 'uniquejobs', job_id, unique_key)
-  return 1
+  return 3
 else
-  return 0
+  return 2
 end
--- a/redis/release_lock.lua
+++ b/redis/release_lock.lua
@@ -4,9 +4,24 @@ local stored_jid = redis.pcall('get', unique_key)
 
 if stored_jid then
   if stored_jid == job_id or stored_jid == '2' then
-    redis.pcall('del', unique_key)
-    redis.pcall('hdel', 'uniquejobs', job_id)
-    return 1
+    if redis.pcall('del', unique_key) then
+      if redis.pcall('hdel', 'uniquejobs', job_id) then
+        -- All should be good
+        return 1
+      else
+        -- hdel has failed
+        return 2
+      end
+    else
+      -- del has failed, lets try hdel anyway.
+      if redis.pcall('hdel', 'uniquejobs', job_id) then
+        -- All should be good anyway ?
+        return 3
+      else
+        -- hdel has failed
+        return 4
+      end
+    end
   else
     return 0
   end
--- a/lib/sidekiq_unique_jobs/lock/until_executed.rb
+++ b/lib/sidekiq_unique_jobs/lock/until_executed.rb
@@ -46,16 +46,16 @@ module SidekiqUniqueJobs
                               argv: [item[JID_KEY], max_lock_time])
         case result
         when 3 # previously 1, was considered a success
-          logger.debug { "successfully locked #{unique_key} for #{max_lock_time} seconds (hsetnx called)" }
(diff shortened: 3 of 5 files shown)


redis/acquire_lock.lua
```diff
@@ -7,13 +7,14 @@ if stored_jid then
   if stored_jid == job_id then
     return 1
   else
+    -- maybe we should do something special in this case?
     return 0
```
mathieujobin, line 11: this is happening a lot for us ...

redis/release_lock.lua
```diff
@@ -4,9 +4,24 @@ local stored_jid = redis.pcall('get', unique_key)
 
 if stored_jid then
   if stored_jid == job_id or stored_jid == '2' then
-    redis.pcall('del', unique_key)
-    redis.pcall('hdel', 'uniquejobs', job_id)
-    return 1
+    if redis.pcall('del', unique_key) then
+      if redis.pcall('hdel', 'uniquejobs', job_id) then
```
mhenrixon, line 8: This will leave a lot of jobs in memory unfortunately.

redis/release_lock.lua
```diff
@@ -4,9 +4,24 @@ local stored_jid = redis.pcall('get', unique_key)
...
+        -- All should be good
+        return 1
+      else
+        -- hdel has failed
+        return 2
+      end
+    else
+      -- del has failed, lets try hdel anyway.
+      if redis.pcall('hdel', 'uniquejobs', job_id) then
```
mhenrixon, line 17: Ah ok you put an extra one here. Seems like unnecessary complexity. Very hard to follow, wouldn't just storing the result in variables be a slicker way of handling this?

mathieujobin, line 17: didn't mean to be concise, smart or pretty or anything like that.
I am keeping functionality 100% one-for-one, I believe.
but adding more return code to help troubleshooting

because the lua scripts returns only 0/1, we don't know exactly what is happening.
with this, I have a better idea, although I dont know enough how this gem works to understand it fully.

i just turn several of my workers to be using `:until_and_while_executing` and looking at the logs, for one given job trying to pick up the lock, it looks like it is trying to steal the lock from the others... https://gist.github.com/mathieujobin/27aa191896624994d536

it does not make much sense to me.

is it possible there is some lock common to a unicity type ?

thanks

mathieujobin, line 17: also, I wasn't sure if `del` and `hdel` could be done in any order regardless of the other one success....

if so, I understand your surprise. as you suggest, I could probably store the result of each and return the proper return code after that

redis/release_lock.lua
```diff
@@ -4,9 +4,24 @@ local stored_jid = redis.pcall('get', unique_key)
...
+        -- All should be good anyway ?
+        return 3
```
mathieujobin, line 19: this case is currently considered a succes, but maybe it should not... 
something we could/should do here?

redis/release_lock.lua
```diff
@@ -4,9 +4,24 @@ local stored_jid = redis.pcall('get', unique_key)
 
 if stored_jid then
   if stored_jid == job_id or stored_jid == '2' then
```
mathieujobin, line 6: what is `2` here?

PR #194, opened 2016-08-31
unique args need to be an array

pboling:
I think the documentation led me astray.  Followed this example for ActiveJob, but that returns a hash, not an array, and my job uniqueness was not enforced.


mhenrixon:
Thank you for the contribution

PR #204, opened 2016-12-30
Increase sleep delay in WhileExecuting#synchronize

dsander:
Sleeping for 100ms instead of 1ms reduces the load both on the sidekiq process as well as redis. Before a few jobs trying to acquire the lock slowed down running jobs significantly by spending almost all of the available cpu time checking for the lock.

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/while_executing.rb
+++ b/lib/sidekiq_unique_jobs/lock/while_executing.rb
@@ -14,7 +14,7 @@ module SidekiqUniqueJobs
 
       def synchronize
         @mutex.lock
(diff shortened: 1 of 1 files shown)


mhenrixon:
Thanks for catching that @dsander! I'll get this merged sometime before the new year.

sbonami:
Psst @mhenrixon, is this the Lunar New Year you're referencing? 😂  (Please merge when you get a chance 😉)

mhenrixon:
Heh I didn't like all the breaking builds but I haven't gotten a chance to fix that yet.

PR #227, opened 2017-06-10
Add activesupport as a dependency

yoshdog:
There are non standard ruby methods called like:
- String#constantize
- Class#class_attribute

When using this gem in non rails apps, we get a no method error for
these methods.

In this change, we add activesupport as a dependency and require the
following modules which contains the missing methods:

```ruby
require 'active_support/core_ext/class/attribute'
require 'active_support/core_ext/string/inflections'
```


mhenrixon:
Thanks for brining it to my attention. Unfortunately I don't want to have active support as a dependency. I'd rather just write some simple ruby code (or copy the parts I need from the AS gem).

yoshdog:
thanks @mhenrixon 

Any chance you can push up the updated version to rubygems?

PR #229, opened 2017-06-13
Use hscan for Util#expire

dmkc:
For large apps, `hgetall` can take forever which will block Redis and suspend any Sidekiq job execution. Use `hscan` instead. 

A question: is it useful to report the `removed_keys`? The issue is that with a huge app collecting all the IDs this will end up cloning `uniquekeys` in RAM on the app server. One of our production apps has a multi-gigabyte `uniquekeys` hash and running `expire` will kill it. If this is useful, I could add something like `Util#silently_expire` with a matching Rake task.


mhenrixon:
I'd love to be able to report the deleted keys, I will get this merged and maybe you could create a second PR for the silently_expire?

dmkc:
👍  Thanks!

PR #232, opened 2017-06-20
Your testing lib is broken and don't permit to test uniqueness of jobs

keysen:
Hello,

I discovered an issue when migrating from 3.0.11 to 4.0.18.

Inside this file: 
>lib/sidekiq_unique_jobs/testing.rb

You decided to do that:
```ruby
          _server.call(worker_class.new, item, queue, redis_pool) do
            call_real(worker_class, item, queue, redis_pool) do
               yield
             end
           end
```

Instead of that (my solution):
```ruby
          call_real(worker_class, item, queue, redis_pool) do
            _server.call(worker_class.new, item, queue, redis_pool) do
              yield
            end
          end
```

It does not make sense because it means in your version you run the server first (the server is the one who performs the job), then you run the client (client pushs the jobs)

It means in details, at first you release the lock then in the client you acquire a lock, but NO, at first you should acquire a lock... and then  after processing the job the lock should be released by the server?! So basically when you want to execute complicated specs where the uniqueness is important we faced of an issue because the lock just don't work. I wasted many days to understand why.

I left some puts in the PR to help you to understand the issue :). But this should of course be removed. 

Otherwise good job for the 4.0.18 it's much better than 3.0.11 :+1: 

**Edit:** sorry for the previous PR (https://github.com/mhenrixon/sidekiq-unique-jobs/pull/231) I was unable to provide you a valid PR without comparing with master and my previous PR was relying on a tag version, so I recreated a new one.

### DIFF
--- a/lib/sidekiq_unique_jobs/client/middleware.rb
+++ b/lib/sidekiq_unique_jobs/client/middleware.rb
@@ -12,6 +12,7 @@ module SidekiqUniqueJobs
       include OptionsWithFallback
 
       def call(worker_class, item, queue, redis_pool = nil)
+        puts "* Client processing: #{worker_class}"
         @worker_class = worker_class_constantize(worker_class)
         @item = item
         @queue = queue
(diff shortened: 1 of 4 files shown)


mhenrixon:
Thank you!

keysen:
Awesome thank you :+1:

PR #237, opened 2017-07-13
Make spec file required files load in deterministic order

jamesr2323:
I was getting uninitialized constant errors because my machine was loading in some other undetermined order


mhenrixon:
I don't see how this would really fix anything other than by chance (the next file builds on top of a previously loaded file just by name). A better fix would be to have the files that break load the needed file.

A better solution in my opinion would be to add explicit requires when necessary (a file needs a dependency in another file). Either change to that or I will close this PR since I haven't had any problems with this before.

mhenrixon:
I’m closing this for now. Check master if this is still an issue. I believe explicit requires from file to file in the case there is a hard dependency would be a much more robust solution.

PR #241, opened 2017-08-20
Better runtime locks

### LINKED ISSUE #230: Version v5.0.5 might have introduced a breaking change in while_executing and was not documented
Hi,

After updating the gem on my server I noticed that the jobs are run differently. It seems that the v5.0.5 and the mutex change from instance to class changed the behaviour and sadly it was not documented.

Previously, when using `while_executing` I was able to limit unique jobs for the same class with passing different unique_args - meaning some jobs actually might have been run in parallel but with different unique arguments.

For example, a certain User could not run multiple Work jobs, but multiple different Users could run their Work job in parallel. For example (jobs in Sidekiq, enqueued simultaneously):

1. Work(unique_args: ['user_1']) - this is a job enqueued by user_1
2. Work(unique_args: ['user_1']) - this is a job enqueued by user_1
3. Work(unique_args: ['user_2']) - this is a job enqueued by user_2

So in previous version of the gem, jobs 1 and 3 would run in parallel (since unique_args values were not unique), and job 2 would wait for job 1 to end.

After the change (and only I'm assuming that) all Work classes are using the same mutex, therefore only one Work class can be run at a time, meaning that in the above example job 3 would wait for job 1 and/or 2 to finish before actually running.

Is there a way to get the previous approach other than using the version before v5.0.5?

### LINKED ISSUE #233: Deadlock using :while_executing?
Hello,

I'm using this library to allow to enqueue multiple jobs but only process 1 job of a time for a certain parameter (in our case a model id).

If I enqueue multiple jobs with no workers running, when I start a worker it deadlocks trying to run multiple jobs of the same parameter at the same time.

image

Any suggestions?

### LINKED ISSUE #243: Trouble with "inline" mode
Hi!  Thanks for your gem!  It's very useful.

When I integrated sidekiq-unique-jobs, we started seeing Sidekiq no longer respecting the inline mode.  It would try to call out to Redis, despite `Sidekiq::Testing.inline!` having been enabled.

I hunted around and eventually found this code snippet in your Gitter chat room:

```
  # https://gitter.im/mhenrixon/sidekiq-unique-jobs?at=579288da32bd01d843a4e2e8
  module SidekiqUniqueJobs
    module Lock
      class UntilExecuted
        def lock(scope)
          if scope.to_sym != :client
            raise ArgumentError, "#{scope} middleware can't #{__method__} #{unique_key}"
          end

          logger.debug { "faked lock #{unique_key} for #{max_lock_time} seconds" }
          true
        end
      end
    end

    module Unlockable
      def unlock_by_key(unique_key, jid, redis_pool = nil)
        logger.debug "faked unlock for #{unique_key}"
        true
      end
    end
  end
```

In the chat room "Mika Hel @mhenrixon" goes on to say:

> Mika Hel @mhenrixon Jul 25 2016 01:29
> @dorkusprime, @wagenet your issues should be mostly resolved in the newest version of the gem. Released yesterday evening

... I am using version `5.0.10` -- which is the latest version released, however, I still have to include that monkey patch.  Is the fix definitely included in the gem?  What am I doing wrong?

Thanks,
Kevin

mhenrixon:
Close #233
Close #230
Close #243


pedrocunha:
hi @mhenrixon are you planning to do a release with this ?

mhenrixon:
Hi @pedrocunha there are a couple of things left to test and I have simply not had the time to do it yet. If you are feeling adventurous you could try the master branch. 

Need to also rewrite the documentation to better explain how the locking is thought to work but between having a baby, working full time and executing a move to another country with my family there isn't much left. 

If people would test the master branch in real applications and test suites that would be very helpful.

kieran:
Any updates on the 5.1.0 release? We have experienced issues wrt `until_executing` in production

Congrats on the baby / move!

PR #248, opened 2017-10-15
Don't skip monkeypatches if ActiveSupport present

dleavitt:
_TLDR: removes two superfluous-seeming checks in core_ext that cause breakage if ActiveSupport is in the require path but not fully loaded._

sidekiq-unique-jobs needs a small number of methods from ActiveSupport. It vendors these methods in case ActiveSupport isn't present.

Currently those methods are only added if ActiveSupport is _nowhere Ruby's require path_. This breaks in the following cases:
- ActiveSupport is being used the application but only a subset of methods are included.
- ActiveSupport is present in the load path but is not being used as part of the application.

This removes the check on whether ActiveSupport is loadable. The solves the above two cases , while still not avoiding re-adding methods in the case that ActiveSupport has already added them.

### DIFF
--- a/lib/sidekiq_unique_jobs/core_ext.rb
+++ b/lib/sidekiq_unique_jobs/core_ext.rb
@@ -1,43 +1,38 @@
 # frozen_string_literal: true
 
-begin
-  require 'active_support/core_ext/hash/keys'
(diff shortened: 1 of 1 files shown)


mhenrixon:
Thank you for the contribution!

PR #262, opened 2018-04-27
Remove unnecessary monkey patches for String

zormandi:
The current stable version of the gem monkey patches a couple of core classes, including String. Unfortunately, the definition of the `camelize` method conflicts with other gems' definition of this same method including Active Support and Backports (see [https://github.com/marcandre/backports/blob/v3.11.3/lib/backports/rails/string.rb](url)). I actually noticed the behavior by requiring backports before sidekiq-unique-jobs and getting a failure of `Wrong constant name 'untilExecuted'`.

The locking mechanism has already been cleaned up in the master branch (although it hasn't been released yet) but the - now unused - monkey patches for String have been left over. I think removing these can spare developers hunting down subtle bugs.

### DIFF
--- a/lib/sidekiq_unique_jobs/core_ext.rb
+++ b/lib/sidekiq_unique_jobs/core_ext.rb
@@ -48,27 +48,3 @@ class Array
     end
   end
 end
-
-class String
-  unless ''.respond_to?(:classify)
-    def classify
-      camelize(sub(/.*\./, ''))
-    end
-  end
-
-  unless ''.respond_to?(:camelize)
-    def camelize(uppercase_first_letter = true)
-      string = self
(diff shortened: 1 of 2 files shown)


zormandi:
Travis build failure seems completely unrelated: `no such file to load -- concurrent/map`. Color me confused.

mhenrixon:
The problem is that removing these monkey patches breaks the gem for people not using active support (which is quite a few). Unfortunately I am not able to use refinements due to jruby support.

brettburley:
@mhenrixon you removed the usage of `classify` in 88ed7a96146799d5b01a550a7cccc937045c419c, so I believe that these monkeypatches are no longer necessary for sidekiq-unique-jobs to work.

PR #265, opened 2018-06-03
Add explicit concurrent-ruby dependency.

brettburley:
The latest master of sidekiq no longer uses the concurrent-ruby dependency.

Current master is failing the `sidekiq_develop` tests as a result.


mhenrixon:
Thank you @brettburley

PR #266, opened 2018-06-05
Fix inverted logic in WhileExecuting lock.

brettburley:
Previously, the lock was being unlocked immediately before the job was yielded to.

@mhenrixon I know it's been a while since you did your refactor here on how locks work -- but please let me know if I'm missing something here -- it looks like the current master unlocks the job immediately after acquiring the lock for the `WhileExecuting` lock.

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/while_executing.rb
+++ b/lib/sidekiq_unique_jobs/lock/while_executing.rb
@@ -17,14 +17,15 @@ module SidekiqUniqueJobs
       # @return [Boolean] report success
       # @raise [SidekiqUniqueJobs::LockTimeout] when lock fails within configured timeout
       def execute(callback)
-        jid = @lock.lock
-        if jid == @item[JID_KEY]
-          callback&.call
-          unlock(:server)
+        locked = @lock.lock(@calculator.lock_timeout)
+
+        if locked
           yield
         else
           fail_with_lock_timeout!
         end
+      ensure
+        callback&.call if locked && unlock(:server)
       end
 
       # Unlock the current item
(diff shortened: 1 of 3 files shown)


mhenrixon:
@brettburley the way it is supposed to work is to allow subsequent jobs to be scheduled but only one to run at the same time. The UntilExecuted locks from when it is pushed by the client until it is executed properly. This part is why a new version hasn't been released yet. I've simply not had any inspiration to work on the computer for some time.

spec/support/sidekiq_meta.rb
```diff
@@ -3,15 +3,14 @@
 RSpec.configure do |config| # rubocop:disable Metrics/BlockLength
   config.before(:each, redis: :mock_redis) do
     require 'mock_redis'
-    mock_redis = MockRedis.new
```
mhenrixon, line 6: Won't this change leak a @redis variable to the outside though?

lib/sidekiq_unique_jobs/lock/while_executing.rb
```diff
@@ -17,21 +17,23 @@ def lock(_scope)
       # @return [Boolean] report success
       # @raise [SidekiqUniqueJobs::LockTimeout] when lock fails within configured timeout
       def execute(callback)
-        jid = @lock.lock
-        if jid == @item[JID_KEY]
-          callback&.call
-          unlock(:server)
+        locked = @lock.lock(@calculator.lock_timeout)
```
mhenrixon, line 20: It is possible the corresponding tests for this file is broken and needs fixing. I haven't had time to go through it all. Wanted to organise the tests and the files better.

mhenrixon:
Thank you for th contribution! These problems should be fixed in master. Recommend you try v6.0.0.beta3 when it is released. 

I’m closing this for now. If you want you can reopen this against v5.x branch. I might do a minor patch on v5 with some minor fixings.

PR #275, opened 2018-06-20
Code smells

mhenrixon:
- Cleanup a lot of code smells
- Break up huge methods into digestible parts
- Rename some stuff
- Remove duplicated (connection/redis) methods and use from module
- Improvements in testing


lib/sidekiq_unique_jobs/scripts.rb
```diff
@@ -7,35 +7,44 @@
...
-    module_function
+    include SidekiqUniqueJobs::Connection
 
-    extend SingleForwardable
-    def_delegators :SidekiqUniqueJobs, :connection, :logger
+    module_function
 
     def call(file_name, redis_pool, options = {})
-      internal_call(file_name, redis_pool, options)
+      execute_script(file_name, redis_pool, options)
```
mathieujobin, line 17: Thank you ! it looks like this is fixing the deprecation warning about redis-namespace

Issue #291, opened 2018-07-15
Making the GEM compatible with Ruby < 2.3


florrain:
Hey @mhenrixon 

I would like to use the latest version in a Ruby 2.1 application. Loading the GEM with that version does not work because of the new safe navigation operator `&.` in the BaseLock class. There's also the use of the `Integer#positive?` that was added in 2.3 that makes the client middleware to explode.

I personally don't think helper methods should prevent the usage of your GEM, would you accept a PR to fix those? And enforcing the Ruby version in the gemspec to be `>= 2`?

If you know of other references to recent Ruby methods that I can add to that PR that would be helpful as well.

Thanks

mhenrixon:
Please go ahead! You are so right in that the helper methods used shouldn’t prohibit the gem from being used. 

Check the Sidekiq requirements though. I think in later Sidekiq versions 2.1 might not be usable.

florrain:
Great will have that PR up hopefully later today or early next week. Good idea, will double check on that 👍

florrain:
Another question for you -- do you have an estimated release date for the final v6?

mhenrixon:
I was hoping to get some people to test it out first and check that everything is working ok. Any time someone suggests everything is working as advertised I'll cut a final v6.0.0.

The gem has been installed a few times now and since I haven't had that many complaints I'd say it is pretty fine for release.

mhenrixon:
I'm closing this for now, ping me on gitter if you need any help with the Ruby 2.1 work.

PR #292, opened 2018-07-15
Ruby < 2.3 compatibility

florrain:
## Context
As discussed in https://github.com/mhenrixon/sidekiq-unique-jobs/issues/291

## Changes
This PR replaces the references to Ruby 2.3 methods to make the GEM compatible with 2.0 and above.

Also added an explicit `required_ruby_version` in the gemspec to be 2.0 and greater. Looking at the README: `Starting from 5.0.0 [...] support for MRI <= 2.1 is dropped`. Is it safe to relax the Ruby requirement? Note that [Sidekiq 4 requires a minimum of Ruby 2.0](https://github.com/mperham/sidekiq/tree/v4.2.10#requirements)

## Testing
I only tested a `:until_executed`, I need to review the rest of the code to confirm the compatibility. The whole test environment is not passing with Ruby 2.1.10 because of incompatibilities with recent dependencies but other than that the core spec files are passing after those fixes.

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/base_lock.rb
+++ b/lib/sidekiq_unique_jobs/lock/base_lock.rb
@@ -73,7 +73,8 @@ module SidekiqUniqueJobs
       end
 
       def callback_safely
-        callback&.call
+        return if callback.nil?
+        callback.call
       rescue StandardError
         log_warn("The lock for #{item[UNIQUE_DIGEST_KEY]} has been released but the #after_unlock callback failed!")
         raise
--- a/lib/sidekiq_unique_jobs/locksmith.rb
+++ b/lib/sidekiq_unique_jobs/locksmith.rb
@@ -84,7 +84,7 @@ module SidekiqUniqueJobs
 
     def grab_token(timeout = nil)
       redis(redis_pool) do |conn|
-        if timeout.nil? || timeout.positive?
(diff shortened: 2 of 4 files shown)


mhenrixon:
The easiest approach is to change `.rubocop.yml` TargetRubyVersion: 2.0 and run rubocop and check for errors. I think you we might have to reconsider all the strings `'MyString'.freeze` to avoid bleeding memory for older gem version. 

It isn't that much work apart from the `.freeze`. We  can leave the frozen string literal in there I think.

mhenrixon:
I put most of the repetitive strings as constants to make this a little easier: https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/constants.rb#L3

Honestly though I mean https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/ ruby 2.2 isn't even supported anymore. Any chance you could just upgrade?

florrain:
Indeed there are more objections thrown by Rubocop after lowering the TargetVersion. Considering the intrusive changes and the incompatibilities with deps, I think I'm going to implement the changes in a fork and then will use your official version as soon as we've finished the upgrade to >= 2.3. Your suggestions are much appreciated, it's going to make it much easier for me, thanks again for the advices!

Going to close this 🔒

PR #314, opened 2018-08-01
Fix replace conflict strategy

RoryDH:
After initially failing to get this feature working, I found that the strategy class wasn't in the hash of strategies.

The other thing this PR addresses is removing the client middleware is loaded on the server. 

The client middleware checks if a lock is in place and enqueues in its absense. Client middleware on the server however is only hit when the server enqueues to itself which happens when sidekiq takes scheduled items and enqueues them on the work queue.

I was trying to use the `:until_executing` lock along with `:replace` strategy, but the server would just take the job and drop in `Client::Middleware#call` because the lock only gets released during the server middleware call.

All ears if the client middleware on the server is necessary for whatever other reason.

All the tests pass without it (bar the test which just checks the presence of it, which I fixed).

### DIFF
--- a/lib/sidekiq_unique_jobs/middleware.rb
+++ b/lib/sidekiq_unique_jobs/middleware.rb
@@ -17,11 +17,6 @@ module SidekiqUniqueJobs
 
     def configure_server_middleware
       Sidekiq.configure_server do |config|
-        config.client_middleware do |chain|
-          require 'sidekiq_unique_jobs/client/middleware'
-          chain.add SidekiqUniqueJobs::Client::Middleware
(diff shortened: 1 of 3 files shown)


mhenrixon:
If you have a look in #315 that should be enough to make the replace work. The thing is that what the replace strategy does is remove the other job from the queue, scheduled queue or retry queue and then also remove the digest and locks. Now the client should be able to push this job again. It has nothing to do with executing the job. That is done from the server process while it is locked from the client push.

RoryDH:
Great thanks for explaining and making those changes!

PR #330, opened 2018-08-24
Convert expiration tine to integer

dareddov:


mhenrixon:
Thanks for the contribution @dareddov! I think it actually used to be `.to_i` at one point. Not sure why it got lost.

PR #343, opened 2018-11-07
Define Config class inside SidekiqUniqueJobs module

### LINKED ISSUE #340: concurrent-ruby 1.1.1 is causing this gem to break
```
identifier Config needs to be constant (NameError)
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/synchronization/abstract_struct.rb:140:in `define_struct_class'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/mutable_struct.rb:213:in `block in define_struct'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/synchronization/mutex_lockable_object.rb:41:in `block in synchronize'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/synchronization/mutex_lockable_object.rb:41:in `synchronize'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/synchronization/mutex_lockable_object.rb:41:in `synchronize'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/mutable_struct.rb:212:in `define_struct'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.1/lib/concurrent/mutable_struct.rb:207:in `new'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/sidekiq-unique-jobs-6.0.6/lib/sidekiq_unique_jobs.rb:45:in `<module:SidekiqUniqueJobs>'
	from /Users/kapil/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/sidekiq-unique-jobs-6.0.6/lib/sidekiq_unique_jobs.rb:40:in `<main>'
```

Slike9:
`Config` class used to be defined in `Concurrent::MutableStruct` which is external, i.e. defined in other library. I guess, it's not good practice to define your code inside a foreign module.
By the way, the pr also fixes #340.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,10 @@ dist: trusty
 sudo: required
 language: ruby
 cache: bundler
+services:
+  - redis-server
 
 before_install:
-  - sudo rm -rf /etc/apt/sources.list.d/rwky-redis.list
-  - sudo add-apt-repository ppa:chris-lea/redis-server -y
-  - sudo apt-get update -qy
-  - sudo apt-get install redis-server
-  - sudo service redis-server start
   - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
   - chmod +x ./cc-test-reporter
 before_script:
--- a/lib/sidekiq_unique_jobs.rb
+++ b/lib/sidekiq_unique_jobs.rb
@@ -43,8 +43,7 @@ module SidekiqUniqueJobs
 
   module_function
 
-  Concurrent::MutableStruct.new(
-    'Config',
+  Config = Concurrent::MutableStruct.new(
     :default_lock_timeout,
     :enabled,
     :unique_prefix,
@@ -54,7 +53,7 @@ module SidekiqUniqueJobs
   # The current configuration (See: {.configure} on how to configure)
   def config
     # Arguments here need to match the definition of the new class (see above)
-    @config ||= Concurrent::MutableStruct::Config.new(
+    @config ||= Config.new(
       0,
       true,
       'uniquejobs',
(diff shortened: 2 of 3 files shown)


mhenrixon:
Thank you @Slike9, the code you suggests I believe will only work for concurrent ruby versions greater than 1.1.1 so we need to enforce that then or provide a fallback to allow both older and newer versions of concurrent ruby to be used. Very please to see this update in concurrent ruby!! I really did not like the way they wanted me to define the config class originally.

Slike9:
@mhenrixon `sidekiq-unique-jobs` requires
```ruby
spec.add_dependency 'concurrent-ruby', '~> 1.0', '>= 1.0.5'
```
`concurrent-ruby` v1.0.0 already supports the api https://github.com/ruby-concurrency/concurrent-ruby/blob/v1.0.0/lib/concurrent/synchronization/abstract_struct.rb#L139. So the fix will work for all now supported versions of `concurrent-ruby`. I've run the tests against  `concurrent-ruby` '1.0.5' and they've passed.

PR #364, opened 2019-01-10
Close #359

### LINKED ISSUE #359: (6.0.7) `uniquejobs:{digest}:AVAILABLE` keys never expire
**Describe the bug**
`uniquejobs:{digest}:AVAILABLE` keys never expire. Given an unbounded set of unique arguments, this will fill up redis. 

**Expected behavior**
In version 6.0.6, these keys don't outlive the execution of their jobs.

**Current behavior**
`AVAILABLE` keys are not cleaned up (introduced in https://github.com/mhenrixon/sidekiq-unique-jobs/pull/354).  If there's a purpose to persisting them, it's not evident (at least in our use case)

**Worker class**

```ruby
class MyWorker
  include Sidekiq::Worker
  sidekiq_options lock: :until_executed
  def perform(args); end

  def self.unique_args(args)
    args
  end
end
```

mhenrixon:
Should fix and close #359

### DIFF
--- a/redis/unlock.lua
+++ b/redis/unlock.lua
@@ -28,5 +28,7 @@ else
   redis.call('DEL', unique_digest)  -- TODO: Legacy support (Remove in v6.1)
 end
 
-return redis.call('LPUSH', available_key, token)
+local count = redis.call('LPUSH', available_key, token)
+redis.call('EXPIRE', available_key, 5)
+return count
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v6.0.8
+
+-  Expire available keys after unlocking ([caused by #352](https://github.com/mhenrixon/sidekiq-unique-jobs/issues/352))
+
 ## v6.0.7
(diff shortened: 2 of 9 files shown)


redis/unlock.lua
```diff
@@ -28,5 +28,7 @@ else
   redis.call('DEL', unique_digest)  -- TODO: Legacy support (Remove in v6.1)
 end
 
-return redis.call('LPUSH', available_key, token)
+local count = redis.call('LPUSH', available_key, token)
+redis.call('EXPIRE', available_key, 5)
```
jwg2s, line 32: Curious as to why you picked 5 seconds here. Worth making a constant?

mhenrixon: No particular reason, do you think 5 is too high? Too low? I mean the key doesn't really do anything except signal to locks that are waiting that the key is now available.

jwg2s: Nah was just curious - sort of a magic number if you're reading that code without context

Issue #357, opened 2018-12-06
Delete all locks button


scicco:
**Is your feature request related to a problem? Please describe.**
I'm still experiencing locks and jobs that stops being scheduled.
I've updated the gem version to 6.0.7 but I'm still seeing the count of uniquejobs raising time after time.
I usually check with this:
`redis-cli --scan --pattern 'uniquejobs:*' | wc -l`

**Describe the solution you'd like**
I'm currently cleaning my environment performing a delete of all keys when the system is "quiet"
`redis-cli --scan --pattern 'uniquejobs:*' | xargs redis-cli DEL`

A "Delete all locks" button inside Sidekiq Unique UI section page would be a nice to have.

**Describe alternatives you've considered**
upgrade of sidekiq version - no luck
upgrade of sidekiq-unique-jobs - no luck
I'm planning to drop sidetiq definitely and use sidekiq-cron instead

**Additional context**

I'm using ruby 2.3.4 version on Ubuntu

My Gemfile:
```
...
gem 'sidekiq-failures', github: 'mhfs/sidekiq-failures'
gem 'sidekiq-unique-jobs', "~> 6.0"
gem 'sidekiq', "~> 5.2.3"
gem 'sidetiq', github: 'tobiassvn/sidetiq'
gem 'sidekiq-status', "~> 0.6"
...
```
Gemfile.lock extract

```
GIT
  remote: https://github.com/mhfs/sidekiq-failures.git
  revision: ade3c0035e8b7df0821a71328fe31dc706226aeb
  specs:
    sidekiq-failures (1.0.0)
      sidekiq (>= 4.0.0)

GIT
  remote: https://github.com/tobiassvn/sidetiq.git
  revision: 261f213d8955ca416625d19dbf4e5cff007ef55b
  specs:
    sidetiq (0.7.2)
      celluloid (>= 0.17.3)
      ice_cube (~> 0.14.0)
      sidekiq (>= 4.1.0)
...
GEM
  remote: https://rubygems.org/
  specs:
    sidekiq (5.2.3)
      connection_pool (~> 2.2, >= 2.2.2)
      rack-protection (>= 1.5.0)
      redis (>= 3.3.5, < 5)
    sidekiq-status (0.8.1)
      chronic_duration
      sidekiq (>= 3.0)
    sidekiq-unique-jobs (6.0.7)
      concurrent-ruby (~> 1.0, >= 1.0.5)
[...]

PR #367, opened 2019-02-07
Add 'Delete All' button to web UI

john-denisov:
Close https://github.com/mhenrixon/sidekiq-unique-jobs/issues/357

I had to to put delete_all action above show action so it would get matched first.
I could have put it as last action if I used HTTP method POST instead of GET. But I assume that there's a reason why POST is not used for regular `delete` form. So I followed the same approach.
  Or we can leave it like that. Maybe it's just me who is used to order all actions by where they belong in `CRUD` acronym 😅


john-denisov:
> Method registered has 26 lines of code (exceeds 25 allowed). Consider refactoring

Can somebody silence this CodeClimate error? 🙂

john-denisov:
@mhenrixon added integration test for `/unique_digests/delete_all` endpoint

mhenrixon:
Thanks @john-denisov, I fixed the style violations in #370 and will merge that as soon as the build passes.

mhenrixon:
Closed with #370

PR #371, opened 2019-02-16
Log job silently complete

tadejm:
Borrows heavily from https://github.com/mhenrixon/sidekiq-unique-jobs/pull/320
with the objective to get this highly useful log shipped to master. 

The problem, as described, is that on a surface a job successfully completes where in fact it wasn't executed at all, due to a lock. 
This commit adds some clarity in the form of a log message explaining that the job did complete silently.

For more info check https://github.com/mhenrixon/sidekiq-unique-jobs/pull/320#issue-206853114

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/until_executed.rb
+++ b/lib/sidekiq_unique_jobs/lock/until_executed.rb
@@ -13,9 +13,12 @@ module SidekiqUniqueJobs
       # Executes in the Sidekiq server process
       # @yield to the worker class perform method
       def execute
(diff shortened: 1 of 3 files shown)


mhenrixon:
Awesome! Thanks for the contribution and the added clarity. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

tadejm:
Thank you for providing this gem, really useful kit! 
Not sure about CI failures–seems like Travis was unable to install all dependencies before running the tests.

Issue #381, opened 2019-04-10
unique_until_and_while_executing without hanging on threads?


JacobEvelyn:
**Is your feature request related to a problem? Please describe.**

We just came across this project—it looks awesome! Thanks so much. We're hoping to use something like `until_and_while_executing` to prevent duplicate jobs from running or getting enqueued too many times, but we're worried about jobs hanging when they get picked up.

**Describe the solution you'd like**

I think the behavior we're hoping for is something like a combination of:

```
lock: :until_executing, on_conflict: :log # So we don't enqueue more than we need.
lock: :while_executing, on_conflict: :reschedule # So threads don't hang and block other jobs.
```

Is something like that possible? There isn't a ton of documentation on `until_and_while_executing` so I'm not sure how it interacts with `on_conflict` in different scenarios. We plan to investigate further and play around with this in library in more depth over the coming days.

**Describe alternatives you've considered**

Like I said, we plan to do more investigation over the coming days. I just wanted to see if this functionality was already built-in or if we should start thinking about workarounds.

Thanks!

mhenrixon:
Thank you for taking the time to add some valuable feedback! You are spot on, this is exactly the direction I want to take the gem. I also want to make some parts of the gem usable directly in the worker.

Would something in https://github.com/mhenrixon/sidekiq-unique-jobs/issues/328 help you?

JacobEvelyn:
Ah, thanks @mhenrixon, I hadn't seen #328. That looks like exactly what we're looking for! Syntax-wise I don't think I've used this gem enough to have a preference among the options you listed there. Is there anything we can do to help with development or testing of this change (with the caveat that this gem and codebase are totally new to us)?

mhenrixon:
The problem I have right now is that I don't know how to write tests for the problems people are experiencing. If anyone could provide a failing test example I can fix the problem but I've only ever managed to run into this from exploratory testing.

I bet you guys are amazing developers and I've reached out to people in the past about pairing on this problem to get it properly fixed but still no takers. It is hard to do it all alone :)

mhenrixon:
@JacobEvelyn if you use `lock_timeout: 0` you don't have any problems. That way the middleware won't wait when it can't acquire a lock. The job will just be dropped, if you want to do something `on_conflict` you could use a specific strategy for that.

mhenrixon:
[...]

Issue #328, opened 2018-08-21
Suggested configuration changes


mhenrixon:
Been thinking a little on the way the worker is configured. There was a suggestion on how to rather than use the `UntilAndWhileExecuting` lock that two separately configured workers with different conflict strategies would be appreciated. The only two ways I can think of doing this would be something like below:

```ruby
### SUGGESTION 1
class ComposedWorker
  include SidekiqUniqueJobs::UniqueWorker
  sidekiq_options queue: :composed, retry: 0, backtrace: true
  lock until_executing: { on_conflict: :reject, expiration: 5.seconds },
       while_executing: { on_conflict: :update, expiration: 10.seconds }
end

### OR 
class ComposedWorker
  include SidekiqUniqueJobs::UniqueWorker
  sidekiq_options queue: :composed, retry: 0, backtrace: true
  lock until_executing: { on_conflict: :reject, expiration: 5.seconds }
  lock while_executing: { on_conflict: :update, expiration: 10.seconds }
end

class UntilExecutingWorker
  include SidekiqUniqueJobs::UniqueWorker
  sidekiq_options queue: :critical, retry: 0, backtrace: true
  lock until_executed: { on_conflict: :log, timeout: 7 }
end

class WhileExecutingWorker
  include SidekiqUniqueJobs::UniqueWorker
  sidekiq_options queue: :default, retry: 0, backtrace: true
  lock while_executing: { on_conflict: :update, expiration: 6.seconds }
end

### SUGGESTION 2

class ComposedWorker
  include Sidekiq::Worker
  sidekiq_options lock: [
    { type: :until_executing, on_conflict: :reject, expiration: 5.seconds },
    { type: :while_executing, on_conflict: :update, expiration: 10.seconds }
  ]
end

class UntilExecutingWorker
  include Sidekiq::Worker
  sidekiq_options lock: { type: :until_executed, on_conflict: :log, timeout: 7 }
end

class WhileExecutingWorker
  include Sidekiq::Worker
  sidekiq_options lock: { type: :while_executing, on_conflict: :update, expiration: 6.seconds }
end
```

[...]

Issue #184, opened 2016-06-06
Feature Request: unique: :around_executing


twetzel:
Hey guys,

I would like to request a new feature:

```ruby
sidekiq_options unique: :around_executing
```
with 2 optional times: (`lock_before_executing` & `lock_after_executing`)
```ruby
sidekiq_options unique: :around_executing, lock_before_executing: 30.minutes, lock_after_executing: 30.min
```

That would be great especially for recurring jobs! 
For Example when I shedule a job for `08:00 am` it would be not possible to shedule another from `07:30 am` to `08:30 am` ... but it would be possible to shedule one at `09:00 am` and another on `10:00 am`.

.. thanks alot ..

Torsten

mhenrixon:
Noted, will think about this for a bit. Should be doable given some consideration

mhenrixon:
Closing — `:around_executing` is effectively what `:until_and_while_executing` provides. It locks around the full lifecycle: enqueue → execution start → execution end.

PR #383, opened 2019-04-15
Provide some configuration DSL for custom Strategies and Locks

mberlanda:
👋 The value proposition of this pull request is to extend the DSL in terms of configuration.

The purpose is to reach something like:

```rb
SidekiqUniqueJobs.configure do |config|
  config.add_lock :some_lock, SomeLock
  config.add_strategy :some_strategy, SomeStrategy

  config.lua_cmd_paths << 'path/to/file.lua' # This may be extracted to another gem
end
```

This could be extended to Connections, Constants, Digests ...

I made this draft/poc out of a new class, but actually some options (`:logger` for example) are already supported in lib/sidekiq_unique_jobs/sidekiq_unique_jobs.rb `SidekiqUniqueJobs::Config`.

Before adding to this `Concurrent::MutableStruct` new elements, I would like make sure that this something suitable for  @mhenrixon 😄 

This may be related to a few open issues:
#184 , #328, #381

Custom Strategies, Locks etc would be available from the code where the gem is included (and we could contribute back to gem once we validated the approach on our projects first)

### DIFF
--- a/lib/sidekiq_unique_jobs.rb
+++ b/lib/sidekiq_unique_jobs.rb
@@ -31,4 +31,6 @@ require "sidekiq_unique_jobs/middleware"
 require "sidekiq_unique_jobs/sidekiq_unique_ext"
 require "sidekiq_unique_jobs/on_conflict"
 
+require "sidekiq_unique_jobs/configuration"
+
 require "sidekiq_unique_jobs/sidekiq_unique_jobs"
--- a/lib/sidekiq_unique_jobs.rb
+++ b/lib/sidekiq_unique_jobs.rb
@@ -31,6 +31,5 @@ require "sidekiq_unique_jobs/middleware"
 require "sidekiq_unique_jobs/sidekiq_unique_ext"
 require "sidekiq_unique_jobs/on_conflict"
 
-require "sidekiq_unique_jobs/configuration"
-
+require "sidekiq_unique_jobs/config"
 require "sidekiq_unique_jobs/sidekiq_unique_jobs"
--- a/lib/sidekiq_unique_jobs/config.rb
+++ b/lib/sidekiq_unique_jobs/config.rb
@@ -10,6 +10,7 @@ module SidekiqUniqueJobs
     :unique_prefix,
     :logger,
     :locks,
+    :strategies,
   )
     DEFAULT_LOCKS = {
       until_and_while_executing: SidekiqUniqueJobs::Lock::UntilAndWhileExecuting,
@@ -21,6 +22,14 @@ module SidekiqUniqueJobs
       while_executing_reject: SidekiqUniqueJobs::Lock::WhileExecutingReject,
     }.freeze
 
+    DEFAULT_STRATEGIES = {
+      log: SidekiqUniqueJobs::OnConflict::Log,
+      raise: SidekiqUniqueJobs::OnConflict::Raise,
+      reject: SidekiqUniqueJobs::OnConflict::Reject,
+      replace: SidekiqUniqueJobs::OnConflict::Replace,
+      reschedule: SidekiqUniqueJobs::OnConflict::Reschedule,
+    }.freeze
+
     class << self
       def default
         new(
@@ -29,6 +38,7 @@ module SidekiqUniqueJobs
           "uniquejobs",
           Sidekiq.logger,
           DEFAULT_LOCKS,
+          DEFAULT_STRATEGIES,
         )
       end
     end
@@ -38,6 +48,13 @@ module SidekiqUniqueJobs
 
       new_locks = locks.dup.merge(name.to_sym => klass).freeze
       self.locks = new_locks
+    end
+
+    def add_strategy(name, klass)
+      raise ArgumentError, "strategy #{name} already defined, please use another name" if strategies.key?(name.to_sym)
+
+      new_strategies = strategies.dup.merge(name.to_sym => klass).freeze
+      self.strategies = new_strategies
     end
   end
 end
--- a/lib/sidekiq_unique_jobs/config.rb
+++ b/lib/sidekiq_unique_jobs/config.rb
@@ -31,6 +31,8 @@ module SidekiqUniqueJobs
     }.freeze
 
     class << self
+      # Returns a default configuration
+      # @return [Concurrent::MutableStruct] a representation of the configuration object
       def default
         new(
           0,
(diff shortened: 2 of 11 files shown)


spec/unit/sidekiq_unique_jobs/configuration_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      it "accepts as many locks as you want" do
+        CustomLock1 = Class.new
+        CustomLock2 = Class.new
+
+        config.add_lock :custom_lock1, CustomLock1
+        config.add_lock :custom_lock2, CustomLock2
+
+        aggregate_failures do
```
mhenrixon: This one is not needed because : https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/spec/spec_helper.rb#L35 

I found it super hard to have a sane test suite without aggregating failures. Didn't like the duplication and turned it on globally.

lib/sidekiq_unique_jobs/configuration.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module SidekiqUniqueJobs
+  # Shared class for dealing with gem configuration
+  #
+  # @author Mauro Berlanda <mauro.berlanda@gmail.com>
+  class Configuration
+    DEFAULT_LOCKS = {
```
mhenrixon: I'd be totally fine with moving this up the chain. Maybe even have this as a global constant under `SidekiqUniqueJobs` directly since it concerns more than just configuration. That way we could remove the other constant `LOCKS` in https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/options_with_fallback.rb#L12.

lib/sidekiq_unique_jobs/configuration.rb
```diff
@@ -0,0 +1,28 @@
...
+      until_expired: SidekiqUniqueJobs::Lock::UntilExpired,
+      until_timeout: SidekiqUniqueJobs::Lock::UntilExpired,
+      while_executing: SidekiqUniqueJobs::Lock::WhileExecuting,
+      while_executing_reject: SidekiqUniqueJobs::Lock::WhileExecutingReject,
+    }.freeze
+
+    def add_lock(name, klass)
+      raise ArgumentError, "Lock #{name} already defined, please use another name" if locks.key?(name.to_sym)
+
+      @locks = locks.dup.tap { |h| h[name.to_sym] = klass }.freeze
```
mhenrixon: This would work too and be a little shorter?

```ruby
@locks = locks.dup.merge(name => klass).freeze 
```

lib/sidekiq_unique_jobs/configuration.rb
mhenrixon: Or possibly put it with the other constants to have all of them in the same place but I am kind of indifferent as long as everything gets wired up correctly.

mberlanda:
hey @mhenrixon thank you very much for your prompt review! I am going to rework my first commit to try to put everything into the `SidekiqUniqueJobs::Config` which exists already.

I will probably drop everything I put into 5991f30 at the end but I wanted to isolate the approach to make it easier to discuss

mberlanda:
I came up with e747b4e for a complete review.

This pr introduces:
* custom (on_conflict) strategies
* custom locks
* a separated configuration object

I managed to increase the overall code coverage.
I still would like to add an integration test to prove that it is actually working what I added.

I also found out a couple of configuration which may be overridden (e.g. `SidekiqUniqueJobs::Digests::`, `Util::DEFAULT_COUNT`, `UtIl::SCAN_PATTERN`) but I would leave the scope as it is to avoid a too large change.

lib/sidekiq_unique_jobs/config.rb
```diff
@@ -0,0 +1,60 @@
...
+    DEFAULT_STRATEGIES = {
+      log: SidekiqUniqueJobs::OnConflict::Log,
+      raise: SidekiqUniqueJobs::OnConflict::Raise,
+      reject: SidekiqUniqueJobs::OnConflict::Reject,
+      replace: SidekiqUniqueJobs::OnConflict::Replace,
+      reschedule: SidekiqUniqueJobs::OnConflict::Reschedule,
+    }.freeze
+
+    class << self
+      def default
```
mhenrixon: The only thing I immediately don't like is `class << self`. I've avoided it like the plague because I find that additional duplication of self very revealing. Especially in large classes where it is difficult to discern what level of nesting I'm at so please do use `self.default`. In my opinion not all duplication are to be avoided :)

I like having the `default` here and using it from the top level though! ❤️

lib/sidekiq_unique_jobs/on_conflict.rb
```diff
@@ -15,17 +15,14 @@ module SidekiqUniqueJobs
...
-    STRATEGIES = {
-      log: OnConflict::Log,
-      raise: OnConflict::Raise,
-      reject: OnConflict::Reject,
-      replace: OnConflict::Replace,
-      reschedule: OnConflict::Reschedule,
-    }.freeze
+    # A convenience method for using the configured strategies
+    def self.strategies
+      SidekiqUniqueJobs.strategies
```
mhenrixon, line 20: I've used delegation a lot in many places but I honestly prefer a defined method and was planning on refactoring in this direction in some places 👍

lib/sidekiq_unique_jobs.rb
```diff
@@ -31,4 +31,5 @@
 require "sidekiq_unique_jobs/sidekiq_unique_ext"
 require "sidekiq_unique_jobs/on_conflict"
 
+require "sidekiq_unique_jobs/config"
```
mhenrixon, line 34: I like the move to a separate file now that it turned complex again! 👍

mberlanda:
@mhenrixon With the latest test pushed and a few doc more, I would consider this PR ready to review. I may add another commit to change the minor version if it is suitable for you but I am not sure about you handle the versioning of the gem.

mhenrixon:
> I may add another commit to change the minor version if it is suitable for you but I am not sure about you handle the versioning of the gem.

When I release a new version I automatically bump to the next at the end of the release as to not forget about it so no need to bump anything. 

You can check the rake task `release` for more info. It isn't perfect but it automates everything for me nearly every time.

mberlanda:
Actually thanks to this change we may define some custom on_conflict strategies specific to our project:

e.g. we use Datadog and we may want to increment a statsd metric instead of logging.
I am pretty sure other users have different vendors and they may want to extend their capabilities

simonbaudry:
Nice improvement 👍

PR #388, opened 2019-04-24
Until and while executing should allow job retry

leandrogoe:
If a job needs to be retried - either because the worker was killed and super fetch retrieved it or the worker raised an exception and was moved to the RetrySet - the `UntilAndWhileExecuting` lock causes the job to be discarded. This is because the lock created by the client middleware no longer exists and the server middleware actively checks that it is still in place and otherwise exits.

In this PR we are changing it so that we do run the worker even after the client lock was wiped out.

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/until_and_while_executing.rb
+++ b/lib/sidekiq_unique_jobs/lock/until_and_while_executing.rb
@@ -16,9 +16,8 @@ module SidekiqUniqueJobs
       # Executes in the Sidekiq server process
       # @yield to the worker class perform method
       def execute
-        return unless locked?
+        unlock if locked?
 
-        unlock
         runtime_lock.execute { yield }
       end
 
--- a/spec/integration/sidekiq_unique_jobs/lock/until_and_while_executing_spec.rb
+++ b/spec/integration/sidekiq_unique_jobs/lock/until_and_while_executing_spec.rb
@@ -36,7 +36,21 @@ RSpec.describe SidekiqUniqueJobs::Lock::UntilAndWhileExecuting, redis: :redis, r
     allow(process_two).to receive(:runtime_lock).and_return(runtime_two)
   end
 
-  it_behaves_like "a lock implementation"
+  it "can be locked" do
+    expect(process_one.lock).to eq(jid_one)
+  end
+
+  context "when process one has locked the job" do
+    before { process_one.lock }
+
+    it "has locked process_one" do
+      expect(process_one).to be_locked
+    end
+
+    it "prevents process_two from locking" do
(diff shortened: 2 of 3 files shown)


spec/integration/sidekiq_unique_jobs/lock/until_and_while_executing_spec.rb
```diff
@@ -36,7 +36,21 @@
...
+
+  context "when process one has locked the job" do
+    before { process_one.lock }
+
+    it "has locked process_one" do
+      expect(process_one).to be_locked
+    end
+
+    it "prevents process_two from locking" do
+      expect(process_two.lock).to eq(nil)
```
mhenrixon, line 51: The problem here is that you actually want to test the `#execute` method, not the lock method. If `process_one` has locked the job should really `process_two` be allowed to execute? I know it is unlikely to happen given the nature of the job but...

lib/sidekiq_unique_jobs/lock/until_and_while_executing.rb
```diff
@@ -16,9 +16,8 @@ class UntilAndWhileExecuting < BaseLock
       # Executes in the Sidekiq server process
       # @yield to the worker class perform method
       def execute
-        return unless locked?
+        unlock if locked?
```
mhenrixon, line 19: it should really be `return unless unlock`. The problem with this PR and the change in logic is that the will run even if another process (job_id) has the lock rendering the first part completely useless.

If you have a look in the locksmith class `unlock` method it actually checks if it is `locked and allowed to unlock. See my comment in the test

leandrogoe: Thanks for the quick feedback! I understand your argument. Since the `WhileExecuting` server lock is still there I thought this was a minor problem since we will still be preventing simultaneous execution.

Excuse my ignorance, but why do you suggest to include this into a lua script?

mhenrixon: There are a couple of changes I have in mind. I am currently working on improving the reliability of the gem, more specifically I am trying to avoid keys being left to rot in Redis.

I'll take your suggestion and improve on it in #387.

mhenrixon: Regarding why put it in the LUA script:

Since we are talking about high concurrency and LUA is transactional we can only really verify the ownership of the lock there.

leandrogoe: Perfect, thanks. Let's close this guy then in favor of #387

leandrogoe:
Closed in favor of #387

PR #435, opened 2019-10-17
remove deprecated/broken OptionsWithFallback#unique_type

zvkemp:
When enqueuing a job, the client middleware continues to view`sidekiq_options unique: :until_executed` as valid, and inserts a lock for the job. However, the Locksmith does not look at the deprecated `unique_type` method (only the `LOCK` key), so when running the lock.lua script, the conditional that should unlock upon completion (`if lock == "until_expired" and ttl then ...`) is never executed.

Discovered on version 6.0.15 (~earlier 6.0.x releases have this issue as well~ edit: appears to have been introduced in 6.0.15), and though 7.0.0.beta2 has been considerably refactored, it still appears to be an issue.

Given that `unique` was deprecated as of 6.0 in favor of `lock`, it seems best to remove it entirely.


mhenrixon:
Thanks a bunch for the contribution! I will backport the fix for v6

PR #439, opened 2019-11-15
Support until executed unlock on raise option

chibicco:
## Issue

The queue will not be unlocked if the sidekiq worker failed to process the job

If sidekiq worker failed to process the job successfully, the queue is not unlocked by default. It would be better if we could add an option to sidekiq_option so that we have the freedom to release the lock if any unexpected error raised during job processing in until_executed mode.

## This PR

Add the unlock_with_exception sidekiq_option which can be used in combination with until_executed.

## Example

Please refer to README for usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,14 @@ Locks from when the client pushes the job to the queue. Will be unlocked when th
 sidekiq_options lock: :until_executed
 ```
 
+#### Unlock with Exception
+
+unlock the queue if error being raised while processing the job in until_executed mode.
+
+```ruby
+sidekiq_options unlock_with_exception: true # Default behavior is false
+```
+
 ### Until Timeout
 
 Locks from when the client pushes the job to the queue. Will be unlocked when the specified timeout has been reached.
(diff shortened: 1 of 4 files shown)


mhenrixon:
Hi @chibicco! 

As much as I appreciate the contribution, this goes against everything I believe in as it will allow duplicate jobs to be added to the queues. Given enough jobs failing with the same argument you might end up with hundreds or even thousands of duplicates. 

I provided functionality for the gem to be extended with locks and strategies from your application code and I believe that would be the way to go. Mike Perham (Contribsys founder and author of Sidekiq) recommends against unlocking the jobs when an exception is raised because it allows for duplicate jobs to be pushed. This is also how Sidekiq Pro/Ent works with unique jobs. I'm not sure I want that code as a default option but I'm happy to provide some way for you to handle it any way you like in your application code so `on_error:` would probably be the way to go. Then you can use your own strategy.

chibicco:
Thank you for the quick response. 

It would be highly appreciated if you can help implement the ```on_error:``` feature. 
Best regards!

tdeo:
Hey @mhenrixon,

I just came accross that same topic, 

> unlocking the jobs when an exception is raised because it allows for duplicate jobs to be pushed

Could you clarify in which situation this can happen and how it would behave differently than `while_executing`? We came across the same use-case and I'd be curious to hear about the actual pitfall with this strategy.

Thanks

mhenrixon:
If we unlock the job and it is retried the lock is gone and a duplicate job can be queued. 

While executing only locks around the job executing and therefor ensures the queue isn't locked.

PR #483, opened 2020-03-29
Use thread-safe digest creation mechanism

zormandi:
We've been getting the following error intermittently (like once a month):

`Digest::Base cannot be directly inherited in Ruby`

It turns out that the Digest::MD5 module used for creating the unique digest for the jobs is not thread-safe. Apparently many people have been having this issue for a long time (e.g. see https://github.com/aws/aws-sdk-ruby/issues/525). Using the OpenSSL module instead is reported to have solved the problem.

### DIFF
--- a/lib/sidekiq_unique_jobs/lock_digest.rb
+++ b/lib/sidekiq_unique_jobs/lock_digest.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'openssl'
+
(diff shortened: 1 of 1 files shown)


mhenrixon:
Wow 😮... Who would have thunk?

Cheers for the contribution. I will get it merged and back-ported sometime today.

mhenrixon:
Released as `v6.0.21` and `v7.0.0.beta14`

PR #497, opened 2020-04-13
[LockType] Allow overwriting default lock type at item level

EmmN:
According to sidekiq documentation, any options from `sidekiq_options` can be overridden per call.
https://github.com/mperham/sidekiq/wiki/Advanced-Options#workers

Based on that, I believe that the way things are implemented right now on `sidekiq-unique-jobs` is not correct, that's why I created this PR.

I spent a couple of hours today trying to understand why the code below was not working as expected. (couldn't find any documentation why you took the decision to implement things like this)

```
class TestWorker
  include Sidekiq::Worker
  sidekiq_options lock: :until_and_while_executing

  def perform
    puts "test"
  end
end

TestWorker.perform_async #=> should use the default lock `until_and_while_executing`
TestWorker.set(:lock => :while_executing).perform_async #=> should use the lock I am specifying `while_executing` but it was actually using `until_and_while_executing`
```

### DIFF
--- a/lib/sidekiq_unique_jobs/options_with_fallback.rb
+++ b/lib/sidekiq_unique_jobs/options_with_fallback.rb
@@ -57,7 +57,7 @@ module SidekiqUniqueJobs
 
     # @return [Symbol]
     def lock_type
-      @lock_type ||= options[LOCK_KEY] || item[LOCK_KEY] || unique_type
+      @lock_type ||= item&.send(:[], LOCK_KEY) || options[LOCK_KEY] || unique_type
     end
 
     def unique_type
--- a/lib/sidekiq_unique_jobs/options_with_fallback.rb
+++ b/lib/sidekiq_unique_jobs/options_with_fallback.rb
@@ -57,7 +57,7 @@ module SidekiqUniqueJobs
 
     # @return [Symbol]
     def lock_type
-      @lock_type ||= item&.send(:[], LOCK_KEY) || options[LOCK_KEY] || unique_type
+      @lock_type ||= item[LOCK_KEY] || options[LOCK_KEY] || unique_type
     end
 
     def unique_type
--- a/spec/unit/sidekiq_unique_jobs/options_with_fallback_spec.rb
+++ b/spec/unit/sidekiq_unique_jobs/options_with_fallback_spec.rb
@@ -46,11 +46,11 @@ RSpec.describe SidekiqUniqueJobs::OptionsWithFallback do
 
     it { is_expected.to eq(nil) }
 
-    context 'when options["unique"] is present' do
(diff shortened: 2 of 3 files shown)


lib/sidekiq_unique_jobs/options_with_fallback.rb
```diff
@@ -57,7 +57,7 @@ def lock_class
 
     # @return [Symbol]
     def lock_type
-      @lock_type ||= options[LOCK_KEY] || item[LOCK_KEY] || unique_type
+      @lock_type ||= item&.send(:[], LOCK_KEY) || options[LOCK_KEY] || unique_type
```
mhenrixon: First of all. Item can never be nil, this code would not be reached without a sidekiq Job hash. Secondly, this code doesn’t really matter. 

The item should contain the lock key from the worker. The original version made sure that in the rare case you change your worker to something else it would pick that up and use it but that might mean changing locks after the lock was created. I can see how this would potentially cause some problems. 

That said, the item can’t be anything except a hash so the send should be removed.

spec/unit/sidekiq_unique_jobs/options_with_fallback_spec.rb
```diff
@@ -46,11 +46,11 @@ def initialize(item, options, worker_class = nil)
 
     it { is_expected.to eq(nil) }
 
-    context 'when options["unique"] is present' do
+    context 'when options["lock"] is present' do
```
mhenrixon, line 49: ❤️

spec/unit/sidekiq_unique_jobs/options_with_fallback_spec.rb
```diff
@@ -126,26 +126,39 @@ def initialize(item, options, worker_class = nil)
...
-      context 'when options["unique"] is present' do
+      context 'when options["lock"] is present' do
         let(:options) { { "lock" => :while_executing } }
 
-        it { is_expected.to eq(SidekiqUniqueJobs::Lock::WhileExecuting) }
+        it { is_expected.to eq(SidekiqUniqueJobs::Lock::UntilExecuted) }
       end
     end
 
+    context 'when options["lock"] is present' do
```
mhenrixon, line 150: ❤️ thanks a ton for the additional coverage. Appreciate it

lib/sidekiq_unique_jobs/options_with_fallback.rb
EmmN: @mhenrixon thanks for reviewing this. I've made a couple changes.

There was a broken spec in which item was defined as `nil` and that is why I've done it initially like this.

Since `item` should never be `nil`, I've changed that spec as well.

Issue #499, opened 2020-05-06
V7 Beta 15 `on_conflict:` with Hash does not work on server


jasonbekolay:
**Describe the bug**
Version 7 Beta 15 does not work on the server side only when given a Hash for `on_conflict:`. Server crashes when a conflict is detected.

**Expected behavior**
Conflict strategy is executed on the server when a conflict is detected.

**Current behavior**
Job crashes with stacktrace:

```
2020-05-06T17:25:16.817Z 67576 TID-owvovowmo WARN: NoMethodError: undefined method `to_sym' for {"client"=>"log", "server"=>"reschedule"}:Hash
Did you mean?  to_s
               to_set
2020-05-06T17:25:16.817Z 67576 TID-owvovowmo WARN: /Users/jasonbekolay/.rvm/gems/ruby-2.6.5@yapp/gems/sidekiq-unique-jobs-7.0.0.beta15/lib/sidekiq_unique_jobs/on_conflict.rb:34:in `find_strategy'
/Users/jasonbekolay/.rvm/gems/ruby-2.6.5@yapp/gems/sidekiq-unique-jobs-7.0.0.beta15/lib/sidekiq_unique_jobs/lock/base_lock.rb:148:in `server_strategy'
/Users/jasonbekolay/.rvm/gems/ruby-2.6.5@yapp/gems/sidekiq-unique-jobs-7.0.0.beta15/lib/sidekiq_unique_jobs/lock/while_executing.rb:40:in `block in execute'
/Users/jasonbekolay/.rvm/gems/ruby-2.6.5@yapp/gems/sidekiq-unique-jobs-7.0.0.beta15/lib/sidekiq_unique_jobs/logging.rb:103:in `block in with_logging_context'
/Users/jasonbekolay/.rvm/gems/ruby-2.6.5@yapp/gems/sidekiq-unique-jobs-7.0.0.beta15/lib/sidekiq_unique_jobs/logging.rb:118:in `block in with_configured_loggers_context'
/Users/jasonbekolay/.rvm/gems/ruby-2.6.5@yapp/gems/sidekiq-5.2.8/lib/sidekiq/logging.rb:48:in `with_context'
```

**Worker class**

```ruby
class TestJob
  include Sidekiq::Worker

  sidekiq_options lock: :until_and_while_executing, on_conflict: { client: :log, server: :reschedule }

  def perform
    print "."
    sleep 1
    puts "."
  end
end
```

**Additional context**
[...]

Issue #495, opened 2020-04-09
V7 - `on_conflict:` no longer accepts a Hash


Matchlighter:
**Describe the bug**
Version 7 Beta 14 does not work when given a Hash for `on_conflict:`.
I'm running on Ruby 2.6.3.

**Expected behavior**
I can provide a Hash to `on_conflict:`

**Current behavior**
When the Conflict occurs, the following Error is thrown:
```
NoMethodError: undefined method `call' for {"client"=>"log", "server"=>"reschedule"}:Hash
<truncated>/sidekiq-unique-jobs-7.0.0.beta14/lib/sidekiq_unique_jobs/lock_config.rb:116:in `on_server_conflict'
<truncated>/sidekiq-unique-jobs-7.0.0.beta14/lib/sidekiq_unique_jobs/lock/base_lock.rb:148:in `server_strategy'
<truncated>/sidekiq-unique-jobs-7.0.0.beta14/lib/sidekiq_unique_jobs/lock/while_executing.rb:40:in `block in execute'
```

**Worker class**

```ruby
class SyncBlueprintWorker
  include Sidekiq::Worker

  sidekiq_options(
    lock: :until_and_while_executing,
    on_conflict: { client: :log, server: :reschedule },
  )
  ...
end
```

**Additional Context**
I've boiled it down to these snippets `on_conflict&.(:[], :server)` & `on_conflict&.(:[], :client)`. I don't understand the need for the `&.(:[], )` syntax (as opposed to just `on_conflict[:client]`) - there's already a guard ensuring it's a Hash.

mhenrixon:
Thanks for the report @Matchlighter, this will be fixed in the next version. I'll hopefully release a new version today.

mhenrixon:
Released as v7.0.15

PR #500, opened 2020-05-06
Return correct conflict strategy if hash uses either strings or symbols as keys.

jasonbekolay:
Potential resolution for #499 

Thank you for your work on this gem!


mhenrixon:
Have a look at https://github.com/mhenrixon/sidekiq-unique-jobs/issues/495 and try out v7.0.16. If it doesn't work for you, please let me know and I will get right to it.

mhenrixon:
Released as `v7.0.0.beta16`

PR #502, opened 2020-05-13
Fix documentation on default_lock_ttl and default_lock_timeout

adeubank:
Thanks for this gem. Cleaning up the documentation as it will throw an exception if not set correctly.

```
NoMethodError: undefined method `lock_ttl=' for #<SidekiqUniqueJobs::Config:0x0000563b42097738>
```

Per this class, https://github.com/mhenrixon/sidekiq-unique-jobs/blob/727bf4c7fcc3abfbace2678ffbf668ac3167b1a2/lib/sidekiq_unique_jobs/config.rb#L6-L7

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@
 - [General Information](#general-information)
 - [Global Configuration](#global-configuration)
   - [debug_lua](#debug_lua)
-  - [lock_timeout](#lock_timeout)
-  - [lock_ttl](#lock_ttl)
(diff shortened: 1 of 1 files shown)


mhenrixon:
Thanks for the reminder! It is actually a bug in the implementation not the documentation. 

I will update the configuration class and all usages when I get a chance.

mhenrixon:
Released as `v7.0.0.beta16`

PR #506, opened 2020-05-19
Deep stringify worker options to account for hash in on_conflict

jasonbekolay:


mhenrixon:
First of all a massive thank you for both PRs! Secondly, would it be possible to get a test that makes this fail unless the deep stringify keys are present? 

I'm asking because obviously there was missing test coverage to begin with and I'd like to improve that :)

spec/sidekiq_unique_jobs/lock_config_spec.rb
```diff
@@ -76,7 +76,7 @@
     it { is_expected.to eq(:log) }
 
     context "when on_conflict is a hash" do
-      let(:on_conflict) { { client: :replace, server: :reschedule } }
+      let(:on_conflict) { { "client" => :replace, "server" => :reschedule } }
```
jasonbekolay: Should match what happens on the client - only keys are stringified

spec/sidekiq_unique_jobs/lock_config_spec.rb
```diff
@@ -88,9 +88,9 @@
     it { is_expected.to eq(:log) }
 
     context "when on_conflict is a hash" do
-      let(:on_conflict) { { client: :replace, server: :reschedule } }
+      let(:on_conflict) { { "client" => "replace", "server" => "reschedule" } }
```
jasonbekolay: Should match what happens on the server - both keys and values are stringified.

spec/sidekiq_unique_jobs/lock_config_spec.rb
mhenrixon: Actually no. In the previous PR I make sure that these are tested properly. This is also not the tests for the file you changed.

In fact this change makes it even less _real_ since most likely the on conflict will be a symbol and my tests here would no longer catch if they are.

spec/sidekiq_unique_jobs/sidekiq_worker_methods_spec.rb
```diff
@@ -61,4 +61,22 @@ def initialize(worker_class)
...
     end
   end
+
+  describe "#worker_options" do
+    subject(:worker_class_constantize) { worker.worker_options }
+
+    let(:worker_class) { UniqueJobOnConflictHash }
+
+    it {
+      is_expected.to eq({
```
mhenrixon: ❤️ great stuff

spec/sidekiq_unique_jobs/sidekiq_worker_methods_spec.rb
```diff
@@ -61,4 +61,22 @@ def initialize(worker_class)
...
+        "lock" => :until_and_while_executing,
+        "on_conflict" => {
```
mhenrixon: I think if the rest of the test suite passes, this test should be good enough! 👍

spec/sidekiq_unique_jobs/lock_config_spec.rb
```diff
@@ -88,9 +88,9 @@
...
 
-      it { is_expected.to eq(:reschedule) }
+      it { is_expected.to eq("reschedule") }
```
mhenrixon: It doesn't matter if this is a string or a symbol because it is always converted to a symbol later anyway : https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/on_conflict.rb#L34

spec/sidekiq_unique_jobs/sidekiq_worker_methods_spec.rb
```diff
@@ -61,4 +61,24 @@ def initialize(worker_class)
...
+
+    let(:worker_class) { UniqueJobOnConflictHash }
+
+    it do
+      expect(worker_options).to eq(
+        {
+          "lock" => :until_and_while_executing,
+          "on_conflict" => {
+            "client" => :log,
+            "server" => :reschedule,
```
mhenrixon, line 76: Yup! This test should be good enough!

spec/sidekiq_unique_jobs/lock_config_spec.rb
jasonbekolay: OK, I've removed the changes to this test. Thanks for the explanation!

PR #507, opened 2020-05-20
Stringify on_conflict hash in Job prepare method

jasonbekolay:
Unfortunately my PR yesterday did not resolve the issue. While we did correct `worker_options`, it turns out that in the client it doesn't follow that code path when there's a conflict. My apologies for missing the mark, I didn't test thoroughly enough.

I was able to determine the root cause of the issue. The `job_hash` is stringified by `Sidekiq.default_worker_options`, and it's just a shallow stringify. That means the `on_conflict` hash still has symbols for keys by the time it gets to `find_strategy`.

The approach I took to resolve is to stringify the hash in `Job#prepare`. I'm happy to change or tweak the approach on your advice.

Thanks for your time recently and the quick release! Hopefully this change fixes this issue for good.

### DIFF
--- a/lib/sidekiq_unique_jobs/job.rb
+++ b/lib/sidekiq_unique_jobs/job.rb
@@ -10,6 +10,7 @@ module SidekiqUniqueJobs
     # Adds timeout, expiration, lock_args, lock_prefix and lock_digest to the sidekiq job hash
     # @return [Hash] the job hash
     def prepare(item)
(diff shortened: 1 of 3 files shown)


mhenrixon:
Released as v7.0.0.beta18!

jasonbekolay:
Thanks for the quick attention and release! Currently running in production and it looks good so far!

PR #465, opened 2020-01-07
Update gemspec: thor

masawo:
Now, the `thor` gem is [version 1.0.1](https://github.com/erikhuda/thor/blob/master/CHANGELOG.md).

On my repository, using sidekiq-unique-jobs, I tried to update `thor` gem, but its version stayed the same.

To resolve this, I modified .gemspec file with reference to an other repository below:
https://github.com/rails/jquery-rails/blob/master/jquery-rails.gemspec#L19

PR #520, opened 2020-06-19
Backport #465 to 6.x series

replaid:
This change has been applied to 7.x, but that series is still in beta. It would be nice for the latest non-beta version of this gem not to be the only one keeping Thor constrained to 0.x versions. Thanks for considering cutting one more 6.x version.


gchan:
@mhenrixon Can a new gem release of v6.x be made so we can make use of this backport? As you can imagine, our application can't resolve dependency conflicts arising from `thor` on v6.0.22 :)

mhenrixon:
Released as v.6.0.23 @gchan @replaid!

PR #545, opened 2020-10-29
Updating README

nrevko:
Adding info on how to turn off uniqueness per job

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@
 - [Usage](#usage)
   - [Finer Control over Uniqueness](#finer-control-over-uniqueness)
   - [After Unlock Callback](#after-unlock-callback)
-  - [Logging](#logging)
+  - [Disable Any Locking Strategy](#disable-any-locking-strategy)
+  - [Logging](#logging)  
   - [Cleanup Dead Locks](#cleanup-dead-locks)
   - [Other Sidekiq gems](#other-sidekiq-gems)
     - [sidekiq-global_id](#sidekiq-global_id)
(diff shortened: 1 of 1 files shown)


mhenrixon:
Hi @nrevko, thank you for the suggestion! 

You can remove `lock: :anything.` Don't think this is a particularly valuable section as you can remove the configuration, or am I missing something?

Also, according to V7 `lock: :lock_name` should be used (not `unique: :lock_name`)

nrevko:
@mhenrixon Just to make sure I understand correctly, adding
`sidekiq_options lock: :anything` will remove any uniqueness constraint for that 1 worker (i.e. disable this gem for that one job), or do I need to write an `anything` lock for that?

mhenrixon:
@nrevko no, I mean just remove `lock: :until_and_while_executing` from the configuration, and you achieve just that. If you add the wrong value, I am sure it will crash somewhere as I use that symbol to find a lock implementation.

mhenrixon:
You can also set `SidekiqUniqueJobs.configuration.enabled = false` which is recommended for your test environment. There is also a more advanced version 

```ruby
SidekiqUniqueJobs.disable! do 
  # perform your code without unique constraint
end
```

Obviously, you can do it without a block to which just toggles `enabled = false` indefinitely.

nrevko:
Thank you for your suggestions! 
I am going to close this. 
We are running on older version of this gem (v5) 🤦‍♀️
Maybe, we'll upgrade by the time v7 is out of beta :)

mhenrixon:
Thanks @nrevko, should hopefully not be long now

PR #581, opened 2021-03-18
Improve compatibility with redis-namespace

### LINKED ISSUE #578: redis-namespace asks to use admistrative commands directly
When Sidekiq configured use redis namespaces like this

```ruby
opts = {
  host:      APP_CONFIG[:redis_host].to_s,
  namespace: REDIS_NAMESPACE
}

Sidekiq.configure_server do |config|
  config.redis = opts
end

Sidekiq.configure_client do |config|
  config.redis = opts
end
```

Deprecated text appears at startup

```
Passing 'info' command to redis as is; administrative commands cannot be effectively namespaced and should be called on the redis connection directly; passthrough has been deprecated and will be removed in redis-namespace 2.0 (at /usr/local/rvm/gems/ruby-2.6.3/gems/sidekiq-unique-jobs-7.0.4/lib/sidekiq_unique_jobs/sidekiq_unique_jobs.rb:191:in `block in fetch_redis_version')
````

### PR DESCRIPTION
**Pull request by Wolfer**
Close #578

Use not namespaced Redis connection to fetch redis_version.

### DIFF
--- a/lib/sidekiq_unique_jobs/sidekiq_unique_jobs.rb
+++ b/lib/sidekiq_unique_jobs/sidekiq_unique_jobs.rb
@@ -188,7 +188,7 @@ module SidekiqUniqueJobs
   # @return [String] a string like `5.0.2`
   #
   def fetch_redis_version
-    redis { |conn| conn.info("server")["redis_version"] }
+    Sidekiq.redis_info["redis_version"]
   end
 
   #

### CODE REVIEW

File: lib/sidekiq_unique_jobs/sidekiq_unique_jobs.rb [hunk 1]
```diff
@@ -188,7 +188,7 @@ def configure(options = {})
   # @return [String] a string like `5.0.2`
   #
   def fetch_redis_version
-    redis { |conn| conn.info("server")["redis_version"] }
+    Sidekiq.redis_info["redis_version"]
```
mhenrixon, line 191: Is this method available on all Sidekiq versions?

Wolfer: since 4.1.
https://github.com/mperham/sidekiq/blame/master/lib/sidekiq.rb#L113

PR #595, opened 2021-04-14
Lock performance

mhenrixon:
While doing some profiling information on the locks, it became apparent that most of the locks' time was spent waiting for concurrent ruby, which was a bug in the wait time for a value. 

**Before:**

image

**After:**

image

Hopefully, this will improve performance a bit as there should be a lot less waiting time.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: 2.7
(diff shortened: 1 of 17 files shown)


ViliusLuneckas:
Hi, anyone here experiencing randomly skipped jobs after this upgrade?

ArturT:
@ViliusLuneckas Do you still see issues?

I tested it so far in development and it seemed to work fine. Didn't test on production yet.

PR #598, opened 2021-04-19
Add option to silence log output after a time out

yboulkaid:
When a lock cannot be acquired, the default behavior is to output a warning to inform the developer.

We ran into a case where we were pushing thousands of jobs to a queue, including many duplicates.
This caused a large number of log lines to be output, drowning our logs in noise:

```
WARN: Timed out after 0s while waiting for primed token (digest: uniquejobs:[...], job_id: [...])
```

This PR introduces a configuration option to allow for disabling this output. We the option defaults
to `false` so that the behavior out of the box is kept the same.

What do you think of this approach/feature?

### DIFF
--- a/lib/sidekiq_unique_jobs/config.rb
+++ b/lib/sidekiq_unique_jobs/config.rb
@@ -18,7 +18,8 @@ module SidekiqUniqueJobs
                                                    :reaper_timeout,
                                                    :lock_info,
                                                    :raise_on_config_error,
-                                                   :current_redis_version)
+                                                   :current_redis_version,
+                                                   :silence_lock_timeout)
 
   #
   # Shared class for dealing with gem configuration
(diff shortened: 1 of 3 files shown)


mhenrixon:
Uff, that reminds me I need a better way of handling this. 

I really appreciate the effort that went into this PR. Don't get me wrong, unless I can come up with a better solution to this problem I will merge it. 

What I have in mind is something more like active support notifications that the gem user can subscribe to. 

This could be used for some instrumentation and debugging as well. Right now, the debugging is done in redis. 

Let me consider this again. I've put it off for quite some time due to not having a clear picture of how it would look. The way you described the problem helped clear out some of that mentally.

yboulkaid:
Thanks for the quick and thorough reply! Yes, an opt-in solution in the style of active support notifications would be really nice 👍 

Let me know if I can help in any way :)

mhenrixon:
@yboulkaid you should see fewer of these log entries in the future. I had a mismatch between the Concurrent ruby waiting time and the waiting time on the blocking command for redis.

yboulkaid:
Great! We will test that out and let you know how it goes, thanks! 🙏  :)

mhenrixon:
Closed by #610 and later #611

PR #599, opened 2021-04-26
Fix recording lock_info

### LINKED ISSUE #589: lock_info set to true but no lock info showing up in web ui
**Describe the bug**
Screen Shot 2021-04-06 at 1 22 03 PM

```ruby
SidekiqUniqueJobs.configure do |config|
  config.debug_lua = true # true for debugging
  config.lock_info = true # true for debugging
  config.max_history = 1000 # keeps n number of changelog entries
  config.reaper = :ruby # also :lua but that will lock while cleaning
  config.reaper_count = 1000 # Reap maximum this many orphaned locks
  config.reaper_interval = 10 # Reap every 10 seconds
  config.reaper_timeout = 5 # Give the reaper 5 seonds to finish
end
```

**Expected behavior**
Expected behavior is to have the lock_info show up in the UI so we can debug jobs

**Current behavior**
Lock info does not show up in the UI.

`To use it turn the following setting on: SidekiqUniqueJobs.config.lock_info = true`- I get this message when I look at the locked job

**Worker class**

Screen Shot 2021-04-06 at 1 23 20 PM

**Additional context**
N/A

mhenrixon:
Close #589

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -57,7 +57,6 @@ detectors:
       - SidekiqUniqueJobs::Orphans::RubyReaper#active?
       - SidekiqUniqueJobs::Orphans::RubyReaper#entries
       - SidekiqUniqueJobs::Redis::Entity#exist?
-      - SidekiqUniqueJobs::SidekiqWorkerMethods#worker_class_constantize
       - SidekiqUniqueJobs::Web::Helpers#cparams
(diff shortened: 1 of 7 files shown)


Laykou:
Thank you very much! When can we expect this to be released? :)

PR #604, opened 2021-05-13
Start new orphan reaper process if orphan reaper is not running

Assa1121:
If a sidekiq process that runs orphan reaper stops unexpectedly and the new process is up within reaper_interval, new orphan reaper won't start.

We've faced a rare issue, when the sidekiq process, that went down, was running both orphan reaper and a job, that needs to be unique. This led to leaving an orphaned lock and the job became stuck.

I believe, this PR can fix this. Each sidekiq process will run reaper ressurector, that starts a new orphan reaper if it looks like there is no reaper process running.


mhenrixon:
This is brilliant, the only thing that might be improved would be that there seems to be a little duplication. 

I haven't had time to properly review the code on my computer but I'll get to it as soon as I can. 

It would be nice to avoid the additional thread but it could be considered an optional feature. 

Thank you

Assa1121:
Thanks a lot for your reply.

I've thought on how an additional thread can be avoided. However, it looks like a running thread is needed if a new reaper should be started on the fly.

Another possible solution can be running an instance of orphan reaper on each sidekiq process and use mutex in redis to make sure only one reaper checks locks during single iteration. Such a solution will cost 1 thread less.

As to duplication, I see what you mean. Will work on that.

lib/sidekiq_unique_jobs.rb
```diff
@@ -41,6 +41,7 @@
 require "sidekiq_unique_jobs/orphans/reaper"
 require "sidekiq_unique_jobs/orphans/observer"
 require "sidekiq_unique_jobs/orphans/manager"
+require "sidekiq_unique_jobs/orphans/reaper_ressurector"
```
mhenrixon: `resurrector` is the correct spelling, this would have to be updated everywhere.

lib/sidekiq_unique_jobs/orphans/manager.rb
```diff
@@ -177,7 +179,7 @@ def enabled?
       # @return [void]
       #
       def register_reaper_process
-        redis { |conn| conn.set(UNIQUE_REAPER, current_timestamp, nx: true, ex: drift_reaper_interval) }
+        redis { |conn| conn.set(UNIQUE_REAPER, current_timestamp, ex: drift_reaper_interval * EXPIRATION_FACTOR) }
```
mhenrixon: 1. The `nx: true` part ensures that the key is only written when it doesn't exist. Why is it being removed?
2. Why are we adding an expiration factor? To clarify what I mean, I added the method drift_reaper_interval to handle the computation. Now the computation is split into 3 locations. The method itself and two places where we duplicate by two. I don't see the need for another variable here and I don't see the need for doubling the time it takes to expire. If we truly wanted to compute something differently it belongs in the `drift_reaper_interval` method but I don't see the need for adding yet another variable to handle the computation.

lib/sidekiq_unique_jobs/orphans/manager.rb
```diff
@@ -187,7 +189,7 @@ def register_reaper_process
       # @return [void]
       #
       def refresh_reaper_mutex
-        redis { |conn| conn.set(UNIQUE_REAPER, current_timestamp, ex: drift_reaper_interval) }
+        redis { |conn| conn.set(UNIQUE_REAPER, current_timestamp, ex: drift_reaper_interval * EXPIRATION_FACTOR) }
```
mhenrixon: Same here regarding `EXPIRATION_FACTOR`

mhenrixon (commented):
Some initial feedback for you :)

lib/sidekiq_unique_jobs/orphans/manager.rb
AlexFlint73: 1. I mistakenly thought that if a new reaper process starts and the key is not updated, resurrector will consider the reaper dead and start a new process. However, I see, that the task is executed right after the process starts and the mutex will be updated. I will revert this change.
2. I faced an issue, when the key expired before the mutex was updated and a duplicate reaping process started. Most likely this happened because I was testing the functionality on low intervals (around 10 seconds) and the `DRIFT_FACTOR` did not increase the interval anyhow. I will retest this on higher intervals.

PR #620, opened 2021-07-01
Ensure `limit` and `timeout` are always numbers

pinkahd:
This should prevent errors where a string is passed
as the `lock_timeout` or `lock_limit` and `.positive?`
is called.

### DIFF
--- a/lib/sidekiq_unique_jobs/lock_config.rb
+++ b/lib/sidekiq_unique_jobs/lock_config.rb
@@ -59,8 +59,8 @@ module SidekiqUniqueJobs
     def initialize(job_hash = {})
       @type        = job_hash[LOCK]&.to_sym
       @worker      = SidekiqUniqueJobs.safe_constantize(job_hash[CLASS])
-      @limit       = job_hash.fetch(LOCK_LIMIT, 1)
-      @timeout     = job_hash.fetch(LOCK_TIMEOUT, 0)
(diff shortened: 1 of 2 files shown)


mhenrixon (approved):
I don't understand how a string can enter there. I'm not super fond of this but if it solves a problem for you I don't mind terribly. 

To my knowledge, you'd have to configure like this:

```ruby
sidekiq_options lock_limit: "2", lock_timeout: "5"
```

I would like to validate against this and raise a helpful message if that's the case.

pinkahd:
> I don't understand how a string can enter there. I'm not super fond of this but if it solves a problem for you I don't mind terribly.
> 
> To my knowledge, you'd have to configure like this:
> 
> ```ruby
> sidekiq_options lock_limit: "2", lock_timeout: "5"
> ```
> 
> I would like to validate against this and raise a helpful message if that's the case.

I'm not 100% sure as well this is our configuration for the job in question 👇🏼 what might happen is the fact that we use Sidekiq 5.2.9 or an older version or Redis 🤔 

```
class SomeWorkerWorker
  sidekiq_options queue: :important, lock: :until_executed, lock_timeout: 1.minute, retry: false
  
  def perform(some_arg_id)
    ...
  end
end
```

But we're still seeing errors such as:

```
undefined method `positive?' for "60":String
```

Screenshot 2021-07-01 at 12 04 54

mhenrixon:
Super weird, thank you for the detailed explanation. I'll cut a new version

Issue #617, opened 2021-06-29
Please try v7.1.0


mhenrixon:
I added an [UPGRADING.md](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/UPGRADING.md#v710) document that I will use for guiding people on how to upgrade. 

A ton of bug fixes went into this release! Even `UntilAndWhileExecuting` should be less problematic now.

Finally a special shoutout to my sponsors. You make all of this work possible and worth it!

ArturT:
After the gem upgrade I noticed this issue (in rails console):

```
3.0.1 :001 > Worker.perform_async("223")
/Users/artur/.rvm/gems/ruby-3.0.1/gems/redis-4.3.1/lib/redis.rb:1291:in `>': no implicit conversion of Integer into Hash (TypeError)
3.0.1 :002 > Worker.perform_async(223)
/Users/artur/.rvm/gems/ruby-3.0.1/gems/redis-4.3.1/lib/redis.rb:1291:in `>': no implicit conversion of Integer into Hash (TypeError)
```

I've checked source code of: `/Users/artur/.rvm/gems/ruby-3.0.1/gems/redis-4.3.1/lib/redis.rb:1291`

```ruby
  def brpoplpush(source, destination, deprecated_timeout = 0, timeout: deprecated_timeout)
    synchronize do |client|
      command = [:brpoplpush, source, destination, timeout]
      require 'pry'; binding.pry # I added it for debugging
      timeout += client.timeout if timeout > 0 # this line is failing on timeout > 0
      client.call_with_timeout(command, timeout)
    end
  end
```

I did 

```
3.0.1 :003 > Worker.perform_async(221)

From: /Users/artur/.rvm/gems/ruby-3.0.1/gems/redis-4.3.1/lib/redis.rb:1292 Redis#brpoplpush:

    1288: def brpoplpush(source, destination, deprecated_timeout = 0, timeout: deprecated_timeout)
    1289:   synchronize do |client|
    1290:     command = [:brpoplpush, source, destination, timeout]
    1291:     require 'pry'; binding.pry
 => 1292:     timeout += client.timeout if timeout > 0
    1293:     client.call_with_timeout(command, timeout)
    1294:   end
    1295: end

[1] pry(#<Redis>)> deprecated_timeout
=> {:timeout=>0}
[2] pry(#<Redis>)> client.timeout
=> 5.0
[3] pry(#<Redis>)> timeout
[...]

PR #624, opened 2021-07-20
Pass lock timeout to primed_async

millerjs:
relates to https://github.com/mhenrixon/sidekiq-unique-jobs/issues/617

There is an issue where an integer `config.timeout` was being
respected in `pop_queued`, but the promise in `primed_async` was only
waiting the `config.ttl` plus the drift.

This means that high latency redis connections will return a `nil`
`primed_jid` and while_executing jobs will get dropped.

The proposed solution is to ensure that `config.timeout` gets passed in
as the `wait` argument to the bound `method(:primed_async)`, which
allows the `Concurrent::Promises` to timeout after the same duration
(plus drift) as the `brpoplpush` call.

The issue triage and approach is documented in:

  - https://github.com/mhenrixon/sidekiq-unique-jobs/issues/617#issuecomment-882826710
  - https://github.com/mhenrixon/sidekiq-unique-jobs/issues/617#issuecomment-883458295
  - https://github.com/mhenrixon/sidekiq-unique-jobs/issues/617#issuecomment-883488756

### DIFF
--- a/lib/sidekiq_unique_jobs/locksmith.rb
+++ b/lib/sidekiq_unique_jobs/locksmith.rb
@@ -102,7 +102,7 @@ module SidekiqUniqueJobs
       raise SidekiqUniqueJobs::InvalidArgument, "#execute needs a block" unless block
 
       redis(redis_pool) do |conn|
-        lock!(conn, method(:primed_async), &block)
+        lock!(conn, method(:primed_async), config.timeout, &block)
       end
     end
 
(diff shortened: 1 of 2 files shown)


mhenrixon (approved):
Stellar research and stellar implementation! I'll get this released as soon as possible!

I have been staring at this problem for so long without understanding what could possibly be wrong.

millerjs:
Thanks and thanks for the great gem!!

mhenrixon:
Released as https://github.com/mhenrixon/sidekiq-unique-jobs/releases/tag/v7.1.4

PR #627, opened 2021-07-28
Fix: UntilAndWhileExecuting

mhenrixon:
@ArturT sorry it took so long, was perplexed and didn't notice the use of ttl instead of timeout.


ArturT:
Thank you. It works fine now. :)

> sorry it took so long

No worry. It's up to you to work on this open-source gem if you feel like it. No pressure.

PR #640, opened 2021-09-27
OnConflict::Replace: yield when lock was achieved

mhenrixon:
Close #629

The replace strategy creates a new lock after deleting the old lock and the old job. After a lock has been achieved we need to allow the worker to yield in that specific scenario.

@dsander can you check if this solves the problem for you?

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -18,7 +18,7 @@ detectors:
     exclude:
       - Sidekiq::JobSet::UniqueExtension#delete_by_value
       - Sidekiq::ScheduledSet::UniqueExtension#delete
-      - SidekiqUniqueJobs::Lock::BaseLock#call_strategy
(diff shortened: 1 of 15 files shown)


dsander:
Thanks for looking into it! Sadly it doesn't work, you can pull in the integration tests I added from this commit https://github.com/projectivetech/sidekiq-unique-jobs/commit/23dc65256084a503220b40560e7eb0db9ce9661d.

When using `on_conflict: :replace` the middleware can not return nil because that tells sidekiq not to enqueue the new job. An approach similar to `:reschedule` might work, but then the caller of `perform_async` still would not get the newly enqueued `jid` and kind of break the Sidekiq API.

mhenrixon:
@dsander the tests you provided passes in this branch now

dsander:
Thank you for fixing it!

Issue #664, opened 2021-12-01
Queue up conflicted jobs and release later


piloos:
In our case we have a third party service which we cannot use in parallel.  Hence, all our jobs using this 3rd party service should be _unique while executing_.  However, each of these jobs has value and must be executed at some point.

In order to accomplish this we implemented two things:
1. We created a custom strategy which queues up the conflicted jobs in a an atomic shared Redis data structure.
2. We use the `after_unlock` callback to push the queued up jobs to Sidekiq again.  We made a small change on the gem to pass the `item` in the `after_unlock` method.

Since our use case seems quite universal, I was wondering if we are over-engineering things here?  Are we overlooking built-in mechanisms to accomplish this use case?  (Note that I am not a fan of _raise/retry_ in Sidekiq because this is dependent on the retry mechanism and timing of Sidekiq, and neither from `reschedule` method since this is imprecise).

Our example code:

```
module UniqueJobStrategies
  class CompanyQueueStrategy < SidekiqUniqueJobs::OnConflict::Strategy
    def call
      company_id = item['args'][1]
      logger.info("Conflict! Holding job for company id #{company_id} in our queue")
      COMPANY_JOB_QUEUE.modify(company_id.to_s) do |array|
        array.push(item)
      end
    end
  end
end
```

```
class TestWorker
  include Sidekiq::Worker

  sidekiq_options lock: :until_executed,
                  lock_info: true,
                  lock_args_method: :lock_args,
                  on_conflict: :company_queue_strategy,
                  lock_ttl: 2.weeks

  def perform(task, *args)
    send(task, *args)
  end

  def self.lock_args(args)
    [ args[1] ]
  end

  def sleep_a_bit(company_id, seconds)
    logger.info("going to sleep for company id #{company_id}...")
    sleep seconds
    logger.info("...waking up for company id #{company_id}")
  end

  def after_unlock(item)
    job = nil
    company_id = item['args'][1]
    logger.info("I was unlocked --> checking for other jobs for company id #{company_id}")
    COMPANY_JOB_QUEUE.modify(company_id.to_s) do |array|
      job = array.shift # take the first one (FIFO)
      array
    end
    if job
      logger.info("Pusing job for company id #{company_id} to Sidekiq")
      job['class'].constantize.set(queue: job['queue']).perform_async(*job['args'])
    end
  end

end
```

Joerg-Seitz:
[...]

PR #665, opened 2021-12-01
Pass `item` in `after_unlock` callback

piloos:
Suggestion to pass `item` in `after_unlock` callback.

Loosely related to issue #664

### DIFF
--- a/lib/sidekiq_unique_jobs/sidekiq_worker_methods.rb
+++ b/lib/sidekiq_unique_jobs/sidekiq_worker_methods.rb
@@ -37,9 +37,9 @@ module SidekiqUniqueJobs
     def after_unlock_hook
       lambda do
         if @worker_class.respond_to?(:after_unlock)
-          @worker_class.after_unlock # instance method in sidekiq v6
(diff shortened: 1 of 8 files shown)


lib/sidekiq_unique_jobs/sidekiq_worker_methods.rb
```diff
@@ -37,9 +37,9 @@ def worker_class
     def after_unlock_hook
       lambda do
         if @worker_class.respond_to?(:after_unlock)
-          @worker_class.after_unlock # instance method in sidekiq v6
+          @worker_class.after_unlock(item) # instance method in sidekiq v6
```
mhenrixon: I love the suggestion, in this case we would likely have to check the arity of the method and only pass the item if the callback method responds to 1 method argument and warn if it does not. 

This would otherwise crash for a lot of users if the method doesn't support an argument?

piloos: You are right.  I wasn't aware of the `arity` check possibility.  I added it.
The PR is blocked on a code climate issue.  Breaking up the method seems to add more complexity...

PR #667, opened 2021-12-02
Fix homepage url

dal-ioki:
The `mhenrixon.github.io` urls (that are shown in rubygems.org) result into `404`


mhenrixon (approved):
Thanks!

PR #679, opened 2022-01-27
Fixing reschedule when using a non default queue

bigzed:
When a job uses a non default queue in it's `sidekiq_options`, the `on_conflict` strategy still enqueues the job into the `default` queue even though, the queue information is available in `item`.

The MR fixes this by calling `set(queue: item['queue'].to_sym)` before `perform_in` in reschedule. It assumes that `item` always contains queue.


mhenrixon:
Hi @bigzed this was a complete oversight by me! Of course, the job should be put on the queue it belongs to.

Could you rebase/merge the main branch and I'll get this released with the next version!

bigzed:
Hello @mhenrixon, the branch is rebased to `main` and the Rubocop offenses are fixed.
Thanks for including it!

mhenrixon:
Released as `v7.1.15` @bigzed

Issue #675, opened 2022-01-11
lock_timeout cannot be nil


thatsanicehat:
**Describe the bug**

In version 7.1.12, setting lock_timeout to `nil` for a given worker to ensure the queue waits indefinitely as described in the docs causes an exception

**Expected behavior**
Setting lock_timeout to `nil` should continuously re-enqueue the job until it can be picked up. 

**Current behavior**

```
NoMethodError: undefined method `+' for nil:NilClass (Most recent call first)

File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 327 in drift
(val + 2).to_f * CLOCK_DRIFT_FACTOR
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 331 in add_drift
val + drift(val)
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 244 in primed_async
.value(add_drift(wait || config.timeout))
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 196 in call
primed_method.call(conn, wait) do |primed_jid|
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 196 in block in lock!
primed_method.call(conn, wait) do |primed_jid|
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 228 in enqueue
yield job_id
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 193 in lock!
enqueue(conn) do |queued_jid|
File /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-unique-jobs-7.1.12/lib/sidekiq_unique_jobs/locksmith.rb line 105 in block in execute
lock!(conn, method(:primed_async), &block)
```

**Worker class**

```ruby
[...]

PR #682, opened 2022-02-14
fix NoMethodError when timeout is nil

oneandonlymike:
This fixes https://github.com/mhenrixon/sidekiq-unique-jobs/issues/675 where the timeout being set to `nil` can cause a NoMethodError

### DIFF
--- a/lib/sidekiq_unique_jobs/locksmith.rb
+++ b/lib/sidekiq_unique_jobs/locksmith.rb
@@ -241,7 +241,7 @@ module SidekiqUniqueJobs
     def primed_async(conn, wait = nil, &block)
       primed_jid = Concurrent::Promises
                    .future(conn) { |red_con| pop_queued(red_con, wait) }
-                   .value(add_drift(wait || config.timeout))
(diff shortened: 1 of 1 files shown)


lib/sidekiq_unique_jobs/locksmith.rb
```diff
@@ -241,7 +241,7 @@ def enqueue(conn)
     def primed_async(conn, wait = nil, &block)
       primed_jid = Concurrent::Promises
                    .future(conn) { |red_con| pop_queued(red_con, wait) }
-                   .value(add_drift(wait || config.timeout))
+                   .value(add_drift(wait || config.timeout || 0))
```
HobbesKrug, line 244: Code looks good, but any idea why `config.timeout` would be nil?

ghost: That is actually expected, the readme suggests it in the example: https://github.com/mhenrixon/sidekiq-unique-jobs#global-configuration

```
config.lock_timeout    = nil   # turn off lock timeout
```

Issue #670, opened 2021-12-09
Skip reaping when queues are too large


francesmcmullin:
**Is your feature request related to a problem? Please describe.**
This is a follow up to [an observation in #663](https://github.com/mhenrixon/sidekiq-unique-jobs/issues/663#issuecomment-985544467), when sidekiq queues are very long (e.g., more than 2,000 jobs) the reaper can take a very long time to run, more than likely hitting its timeout (depending on config).

**Describe the solution you'd like**
For applications with traffic/activity spikes, sidekiq queues can get very full, but they are eventually emptied. It's fine if reaping is postponed until queues are lower and there's more slack in the system.

**Describe alternatives you've considered**
We could play with the reaper timeout and count settings to try and find a balance such that reaping can continue even under heavy load, but for our use case it's not a problem if reaping is delayed during heavy activity.

**Additional context**
All my observations relate to the ruby reaper, it is possible that the lua reaper has no trouble with big queues

mhenrixon:
Thank you for the issue, I'll do some thinking on this but I will likely not be able to achieve much until after Christmas. 

I am moving my family across Germany on Monday.

francesmcmullin:
Of course! You've helped a huge amount already. I hope the move goes really well. Thank you =)

PR #690, opened 2022-03-11
Abort Ruby Reaper when sidekiq queues are full

francesmcmullin:
This is an attempt to resolve https://github.com/mhenrixon/sidekiq-unique-jobs/issues/670 .

This change is quite blunt - it silently returns from the ruby reaper if queues are longer than an arbitrary number (1,000 jobs). I'm conscious this may not be ideal for all use cases, but there are a lot of settings already. In my experience, the performance of the reaper degrades so much in these circumstances, it's not worth running, and it will occupy resources you could be using to clear that queue!

If there's anything I missed or further changes from a stylistic or architectural perspective that might help, just let me know and I'll be happy to work on this some more.

### DIFF
--- a/lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
+++ b/lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
@@ -15,6 +15,9 @@ module SidekiqUniqueJobs
       # @return [String] the suffix for :RUN locks
       RUN_SUFFIX = ":RUN"
       #
+      # @return [Integer] the maximum combined length of sidekiq queues for running the reaper
+      MAX_QUEUE_LENGTH = 1000
+      #
       # @!attribute [r] digests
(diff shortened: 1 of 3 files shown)


lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
```diff
@@ -15,6 +15,9 @@ class RubyReaper < Reaper
       # @return [String] the suffix for :RUN locks
       RUN_SUFFIX = ":RUN"
       #
+      # @return [Integer] the maximum combined length of sidekiq queues for running the reaper
+      MAX_QUEUE_LENGTH = 1000
```
mhenrixon, line 19: If we wanted to make this a little smarter, we could make it a configuration for the reaper. Much like `reaper_timeout` and friends. Perhaps a `reaper_max_queue_size`?

PR #709, opened 2022-05-04
fix: re:lock until_executing on worker failure

### LINKED ISSUE #394: Failed jobs waiting to be retried are not considered when fetching uniqueness
**Describe the bug**
If a job fails, and is waiting to be retried, adding a new job with similar arguments should trigger the conflict resolution mechanism.

I have an job which - based on its arguments - calculates and saves an expensive operation to the database, that I would not like to enqueue multiple times (since it's output is only based on the input, and I want to make the calculation as few times as possible...), but if it is already being executed, allow it t be enqueued. 
Curently I am using this config:

```
    sidekiq_options lock: :until_executing, on_conflict: :replace

   # only use the id for uniqueness checking, and ignore the actual data to be aggregated
    def self.unique_args(args)
      [ args[0] ]
    end

```

`until_executed` is not okay due to the requirement for it to be reexecuted if it is enqueued while it is already running.

Now the issue here is that the lock is removed when the job is scheduled to be run, but it is not readded if it fails.

**Expected behavior**
The system should re-add the lock if the job fails, and trigger the deduplication mechanism if a similar job has been enqueued in the meantime. If you think about it this is not very different than adding a job to the end of the enqueued jobs list, other than it contains some meta-information about the failure, and how many times it was tried.
Not only does it get duplicated, but the execution order is undeterministic, sometimes the job that is scheduled later will be executed first, and when the job from the retry queue is rescheduled, it is executed as well, overwriting the fresh data with old one.
 
**Current behavior**
Jobs get duplicated

mhenrixon:
Closes #394

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/until_executing.rb
+++ b/lib/sidekiq_unique_jobs/lock/until_executing.rb
@@ -33,6 +33,9 @@ module SidekiqUniqueJobs
       def execute
         callback_safely if locksmith.unlock
         yield
+      rescue StandardError => ex
+        reflect(:execution_failed, item, ex)
+        locksmith.lock(wait: 1)
       end
     end
   end
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -21,8 +21,10 @@ SidekiqUniqueJobs.reflect do |on|
   end
 
   # This means your code broke and we caught the execption to provide this reflection for you. It allows your to gather metrics and details about the error. Those details allow you to act on it as you see fit.
-  on.execution_failed do |job_hash|
-    logger.warn(job_hash.merge(message: "Execution failed"))
+  on.execution_failed do |job_hash, exception = nil|
+    message = "Execution failed"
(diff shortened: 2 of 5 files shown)


lib/sidekiq_unique_jobs/lock/until_executing.rb
```diff
@@ -33,6 +33,9 @@ def lock(&block)
       def execute
         callback_safely if locksmith.unlock
         yield
+      rescue StandardError => ex
+        reflect(:execution_failed, item, ex)
+        locksmith.lock(wait: 1)
```
axos88, line 38: Shouldn't the exception be rethrown here?

mhenrixon: Indeed, sorry! Fixed in #712

mhenrixon: @axos88 relesed as v7.1.23 I believe

PR #714, opened 2022-06-02
Fix Sidekiq::Worker.clear_all override not being applied

dsander:
Since the method is defined as a singleton the testing override needs to
be prepended to the singleton class.

### DIFF
--- a/lib/sidekiq_unique_jobs/testing.rb
+++ b/lib/sidekiq_unique_jobs/testing.rb
@@ -91,15 +91,6 @@ module Sidekiq
       end
 
       #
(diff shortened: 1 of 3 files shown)


dsander:
This in build on top of https://github.com/mhenrixon/sidekiq-unique-jobs/pull/713 because I didn't want to deal with merge conflicts 😄 I'll rebase this branch if you accept the first PR.

mhenrixon (approved):

mhenrixon:
Could you try and fix the broken tests and such so that we can get this merged @dsander?

PR #715, opened 2022-06-09
Chore: Sidekiq 6.5 compatibility

mhenrixon:


lib/sidekiq_unique_jobs/middleware/server.rb
```diff
@@ -6,6 +6,8 @@ module Middleware
     #
     # @author Mikael Henriksson <mikael@mhenrixon.com>
     class Server
+      include Sidekiq::ClientMiddleware if defined?(Sidekiq::ServerMiddleware)
```
KieranP, line 9: @mhenrixon Shouldn't this include ServerMiddleware?

mhenrixon: Fixed it now

PR #719, opened 2022-06-30
Do not run reaper if registration failed

francesmcmullin:
I might be misunderstanding the intended purpose of the manager process, but it seems like the reaper should run if registration was successful

### DIFF
--- a/lib/sidekiq_unique_jobs/orphans/manager.rb
+++ b/lib/sidekiq_unique_jobs/orphans/manager.rb
@@ -39,7 +39,9 @@ module SidekiqUniqueJobs
         self.task = test_task || default_task
 
         with_logging_context do
-          register_reaper_process
+          got_lock = register_reaper_process
+          return unless got_lock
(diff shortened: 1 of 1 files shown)


lib/sidekiq_unique_jobs/orphans/manager.rb
```diff
@@ -39,7 +39,9 @@ def start(test_task = nil) # rubocop:disable
         self.task = test_task || default_task
 
         with_logging_context do
-          register_reaper_process
+          got_lock = register_reaper_process
```
mhenrixon, line 42: If you have a look at the top of the method, we check if a manager has been registered already so unless I am completely missing something (happens more often than I'd like to admit); this shouldn't be necessary?

francesmcmullin: Ahh, yes, thank you for explaining! I'll just close this one =)

PR #720, opened 2022-06-30
Avoid skipping ranges when looping through queues

francesmcmullin:
When queues are very full and/or very active, the ruby reaper can't give absolute guarantees because it doesn't run as a single atomic action, but we can eliminate some edge cases.

In this instance, the reaper anticipates that a queue might *empty*, from the front, in which case ranges need to be adjusted downward to ensure every job is checked. However, in the case where the queue *grows* (at the back), this calculation of "next_page - (initial_size - current_size))" means *skipping over* some jobs.

### DIFF
--- a/lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
+++ b/lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
@@ -229,6 +229,7 @@ module SidekiqUniqueJobs
 
         loop do
           range_start = (page * page_size) - deleted_size
+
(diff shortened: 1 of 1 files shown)


mhenrixon:
Rebasing with master will give you a green build on the first two PRs.

mhenrixon (approved):

mhenrixon:
Whatever is left is all on you 😋

francesmcmullin:
Green now 🎉  🟢  Thanks for the nudge ^_^

Issue #684, opened 2022-02-15
10X increase in read ops/sec after upgrade from v6 to v7


jkhulme:
**Describe the bug**
We upgraded from 6.0.24 to 7.1.12 (alongside upgrade sidekiq from 6.1.2 to 6.4.0)

Before the upgrade we were seeing ~300 reads/sec, and ~300 writes/sec
After the upgrade we are seeing ~3000 reads/sec, and still ~300 reads/sec

This is a very big increase to see happen, and is proving to be quite expensive with our redis provider

**Expected behavior**
Not to see a 10x increase in reads/sec

**Current behavior**
We are seeing 10x increase in reads/sec

**Worker class**
n/a

**Additional context**
Please let us know if there is any other info we can provide

mhenrixon:
Sorry about the trouble, you can tweak the reaper settings to cut the reads a bit. 

Most of the work (apart from the reaper) is done in Lua so shouldn't cause too much trouble.

francesmcmullin:
Hi 👋🏻  For context, @jkhulme  and I both work on the same application which depends on your gem (thank you!). We have different deployments with different scales, but a lot of our usage patterns look similar.

I took some time today to examine our Redis Reads and I believe you're right that the largest increase is coming from the reaper. It fails while our sidekiq queues are full, but once they clear out we could have 300,000 locks to clean out (@jkhulme  may have millions). Unfortunately I don't expect reaping less frequently to help - in the scenario where we have 300k locks to clear, it's going to cost a lot of reads whether we clean 1k every 30 seconds or 10k every 5 minutes.

It's possible that switching to the Lua reaper would fix a lot of our problems, but I am concerned about its blocking behaviour, especially when we can build up such a large number of locks to be reaped.

Currently, when the reaper is iterating over a list of digests to find orphans, it performs the following operations *per digest*:
- ZSCAN to check if the digest is scheduled
- ZSCAN to check if the digest is retried
- SSCAN to find sidekiq queues
- for each sidekiq queue: (for us, this is 3)
  - LLEN for initial size
  - LRANGE (queue size / 50) times (if it's more than 1k the reaper will be too slow and fail anyway)
  - LLEN again if the queue had any entries (queue size / 50) times
- SSCAN of sidekiq processes (this is usually just 1 for us, but I think @jkhulme  has lots)
- for each sidekiq process: (pipelined, so possibly counts as just one read)
  - EXISTS
  - HGETALL to find workers
 
[...]

PR #721, opened 2022-06-30
Move until_expired digests to separate zset

francesmcmullin:
This is another potential solution to #684 . Although for our specific use case, we've moved away from sidekiq unique jobs for high traffic until_expired jobs, this may still be helpful for other users.

This is quite a large change, I appreciate I may need to improve my Lua, or make different refactorings.

In simple terms, the idea is to treat `until_expired` digests separately to the others. For all other digests, cleanup looks like this:
1. The job is locked and the digest is saved. Possible when enqueueing, or at another time
2. Something else happens, triggering lock cleanup - this can be the job finishing, or being picked off the queue for execution, all keys relating to the lock and the digest are deleted together.
3. In rare cases, if something goes badly wrong above, the reaper will clean up the digest and related keys.

But for `until_expired`, cleanup looks like this:
1. The job is locked and the digest is saved, usually when enqueueing
2. The expiry time is reached, and all keys relating to the lock are deleted by redis's native expiry system, because they were all set with a TTL
4. The digest, as a member of a ZSET, *cannot have an expiry*, so it *must be cleaned up by the reaper*

For this reason, if you use a *lot* of `until_expired` jobs, the reaper can become very slow and expensive. My proposed solution here is:
- Do not save `until_expired` digests with the others, put them in their own separate ZSET, where the value (score) is *the expiry time*
- When the reaper ones, delete by digest from the separate `until_expired` ZSET based on the score, skip the other (expensive) reaping checks.


mhenrixon:
This is a great suggestion, can't believe I didn't think about this before. Would love to see a green test suite here :)

mhenrixon:
I'll hold off on merging this for a few days because I need to consider how this affects an upgrade. In your case (if you are still using `until_expired`) you'd end up with quite a lot of keys that the reaper must clean up. 

I wonder what the damage would be for people upgrading and what version to create next.

Issue #637, opened 2021-09-15
Memory bloat / dangling keys / reaper not cleaning orphans


sharq1:
**Describe the bug**
The `uniquejobs:...` keys do not get removed from Redis, thus over few months it took almost all of our Redis memory (500MB). See how it grew over 9 weeks:
![image](https://user-images.githubusercontent.com/1241239/133391017-6e61b73b-300b-4248-9579-99bc10b04dc3.png)

I have just manually removed all keys matching `uniquejobs:*` and it released nearly all the memory that was being used:
![image](https://user-images.githubusercontent.com/1241239/133391579-bc5c1298-446a-42d1-91db-5ffd1527da1a.png)

**Expected behavior**
No `uniquejobs` keys are left dangling in Redis once there are no enqueued jobs.

**Current behavior**
Lots of dangling keys are left in Redis.

**Additional context**
We were on version 7.0.1 for a while and switched to 7.1.5 just days ago.

Here's our Sidekiq initilizer:
```ruby
Sidekiq.configure_server do |config|
  config.client_middleware do |chain|
    chain.add SidekiqUniqueJobs::Middleware::Client
  end

  config.server_middleware do |chain|
    chain.add SidekiqUniqueJobs::Middleware::Server
  end

  SidekiqUniqueJobs::Server.configure(config)
end

Sidekiq.configure_client do |config|
  config.client_middleware do |chain|
    chain.add SidekiqUniqueJobs::Middleware::Client
  end
end
```

Some dangling keys found when no jobs were enqueued:
```
[...]

PR #724, opened 2022-07-01
Ensure batch delete removes expiring locks

francesmcmullin:
Hi @mhenrixon thanks so much for the positive response to #721 

Unfortunately I missed a spot! 😅  In truth, I think it's unnecessary to use batch delete when removing expiring locks - it would be enough to just remove the digests from the ZSET because all other elements are expired by redis, but since I *did* use batch delete - it's really important that batch delete removes from this new ZSET!

Previously, I only updated `lock.lua` to put these digests into a separate ZSET because that's what the locksmith calls, but when trying to verify this fix in tests, I found `lock.rb` needed to be updated as well, so that's here too.

### DIFF
--- a/spec/sidekiq_unique_jobs/orphans/ruby_reaper_spec.rb
+++ b/spec/sidekiq_unique_jobs/orphans/ruby_reaper_spec.rb
@@ -135,5 +135,28 @@ RSpec.describe SidekiqUniqueJobs::Orphans::RubyReaper do
         expect(service).not_to have_received(:orphans)
       end
     end
+
+    context "when a lock is until_expired" do
+      let(:lock_info) do
+        {
+          "job_id" => job_id,
+          "limit" => 1,
+          "lock" => :until_expired,
+          "time" => now_f,
+          "timeout" => nil,
+          "ttl" => 1,
+          "lock_args" => [],
+          "worker" => "MyUniqueJob",
+        }
+      end
+
+      it "clears the lock" do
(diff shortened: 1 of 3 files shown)


spec/sidekiq_unique_jobs/orphans/ruby_reaper_spec.rb
```diff
@@ -135,5 +135,28 @@
...
+          "timeout" => nil,
+          "ttl" => 1,
+          "lock_args" => [],
+          "worker" => "MyUniqueJob",
+        }
+      end
+
+      it "clears the lock" do
+        expect(redis { |conn| conn.zcard(SidekiqUniqueJobs::EXPIRING_DIGESTS) }).to eq 1
+        sleep 2
```
francesmcmullin, line 155: This spec passes for me locally, but I don't think a sleep this long is really appropriate even if it did work consistently. It's also not really the right place for a spec which is curious about the interaction between both the reaper and batch delete. It's here to demonstrate the issue, but I'd love your thoughts about a better location or tactic for testing this

mhenrixon: A sleep of two seconds would increase the test suite time with more than 50% (on my machine). Seems like something else is off here, I am without a computer until Monday but I'll check then.

fotinakis:
Hey friends, what's the status on this PR? I hear this will help solve our rollout issues with using :until_expired at heavy scale (https://github.com/mhenrixon/sidekiq-unique-jobs/issues/637#issuecomment-1180476292).

I'd be happy to run against this branch for a large-scale test if that'd be helpful, once it's ready and tests passing.

PR #737, opened 2022-10-26
fix(after_unlock):  fix regression from #707

adamcreekroad:
# Background

#707 introduced a regression where the instance level `after_unlock` callback is no longer being called. This was due to the change of using `item[CLASS]` instead of `worker_class` to set `job_class` in `SidekiqUniqueJobs::Middelware#call`. `item[CLASS]` is a string, where as `worker_class` can be either an instance, or the class itself. For the instance level `after_unlock` to work, we need `job_class` to be an instance of the worker.

This fixes the issue by switching back to setting `job_class` to `worker_class`

# Changelog

- Use worker_class for self.job_class in Middleware#call


mhenrixon:
Hi @adamcreekroad I updated your branch with the latest changes to make sure the CI results are fair to your change. 

Please fix any issues remaining and I'll merge and release your PR as well. Thank you for the contribution.

PR #751, opened 2023-01-09
feat: sidekiq v7 support

mhenrixon:
This means I have to drop compatibility with older sidekiq versions. It just won't be possible to maintain both due to the change in redis api.


rcaught:
Thank you for doing this!

PR #756, opened 2023-02-14
Fix active worker detection by using correct keys

dodo121:
Hello,
We are using this gem for some of our products at [marketer.tech](https://www.marketer.tech/).

**The problem**:
Recently we've spotted an issue with one our workers. The lock was not working and the same cron job was processed multiple times. After some investigation, we noticed that the reaper removes active locks which is incorrect behavior.

**Root cause of the issue**
It turns out that the `active?` method does not find any active workers because of the naming issue. Sidekiq sets the keys with "#{key}:work" instead of "#{key}:workers". The same happens to Lua reaper.

**Solution:**
Rename keys from `:workers` to `:work` since that's the correct name set in Redis:
Zrzut ekranu 2023-02-14 o 13 04 25

**QA**:
We tested those changes with our app and both reapers (Ruby & Lua) no longer remove active locks.

### DIFF
--- a/lib/sidekiq_unique_jobs/lua/shared/_find_digest_in_process_set.lua
+++ b/lib/sidekiq_unique_jobs/lua/shared/_find_digest_in_process_set.lua
@@ -15,7 +15,7 @@ local function find_digest_in_process_set(digest, threshold)
     log_debug("Found number of processes:", #processes, "next cursor:", next_process_cursor)
 
     for _, process in ipairs(processes) do
-      local workers_key = process .. ":workers"
+      local workers_key = process .. ":work"
       log_debug("searching in process set:", process,
                 "for digest:", digest,
                 "cursor:", process_cursor)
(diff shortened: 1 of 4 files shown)


mhenrixon:
Quick question: which Sidekiq and SidekiqUniqueJobs versions are you currently using? I want to make sure this gets backported if it concerns v7.

dodo121:
> Quick question: which Sidekiq and SidekiqUniqueJobs versions are you currently using? I want to make sure this gets backported if it concerns v7.

sidekiq-unique-jobs (8.0.0)
sidekiq (7.0.3)

mhenrixon (approved):

mhenrixon:
Releaased as 8.0.1 and should be available in a few minutes. I know rubygems are cached pretty hard so sometimes takes a minute.

dsander:
We are running `sidekiq-unique-jobs` with Sidekiq 6.5.8, took me a bit to figure out why some jobs were running in a loop 😄. It would be awesome if we could get this backported to v7 of sidekiq-unique-jobs

PR #763, opened 2023-03-09
Hide lock info debug suggestion on lock page if it's already enabled.

JeremiahChurch:
Super small change to hide the suggestion to turn the lock_info config on if the config value is already enabled.

potentially save some head scratching if you have the config enabled but are still seeing that message when you don't have lock info for whatever reason


mhenrixon (approved):
Perhaps this page shouldn't even be available unless the setting is turned on but then it might not be discovered.

Thanks for your contribution.

PR #764, opened 2023-03-13
fix rubocop

JeremiahChurch:
cleanup for #762 & #763

replace `match_array([])` with `eq([])` and rubocop -A run through for the rest.

only one spec had an issue with either of those other matchers - left a rubocop disable on it (digests_spec#132)


mhenrixon (approved):

mhenrixon:
Thank you for the contributions

PR #765, opened 2023-03-13
update debug_lua check to match passed in arg type

JeremiahChurch:
I believe this was introduced with `SidekiqUniqueJobs::Script::Caller::normalize_argv`

The lua scripts check for a value of `"true"` for `debug_lua` but `normalize_arg` changes that to integer 1 that looks to be typed to a string when it gets to the lua script (I don't know lua in any way so this could very well be wrong)

```
def normalize_argv(argv)
  argv.each_with_index do |item, index|
    case item
    when FalseClass, NilClass
      argv[index] = 0
    when TrueClass
      argv[index] = 1
    end
  end
end
```

also remove unused `toboolean` in lua scripts

I have no idea how to add a test for this.

### DIFF
--- a/lib/sidekiq_unique_jobs/lua/delete.lua
+++ b/lib/sidekiq_unique_jobs/lua/delete.lua
@@ -17,7 +17,7 @@ local limit        = tonumber(ARGV[4])
 
 --------  BEGIN injected arguments --------
 local current_time = tonumber(ARGV[5])
-local debug_lua    = tostring(ARGV[6]) == "true"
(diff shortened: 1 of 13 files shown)


mhenrixon (approved):

mhenrixon:
This looks sane; I don't know if testing is possible since it happens inside redis. One would have to subscribe to the log file of redis. 

The change you are referring to was done because of `redis-client` not supporting additional data types. This was one of the hardest parts of sidekiq 7 compatibility.

PR #778, opened 2023-04-27
Because `replace` is a client strategy, it should only remove client locks aka queue locks.

bigzed:
I just tried to use `replace` with `until_and_while_executing` and noticed, that when I spam `TheLockedJob.perform_async(1)` it would replace the job in the queue and also delete the runtime lock of the already running job leading to two jobs running at the same time.

The change results in the job correctly being replaced on the queue without removing existing runtime locks for the job.

### DIFF
--- a/lib/sidekiq_unique_jobs/digests.rb
+++ b/lib/sidekiq_unique_jobs/digests.rb
@@ -7,6 +7,16 @@ module SidekiqUniqueJobs
   # @author Mikael Henriksson <mikael@mhenrixon.com>
   #
   class Digests < Redis::SortedSet
+    #
+    # @return [Integer] the number of matches to return by default
+    DEFAULT_COUNT = 1_000
+    #
+    # @return [String] the default pattern to use for matching
+    SCAN_PATTERN  = "*"
+    #
+    # @return [Array(String, String, String, String)] The empty runtime or queuetime keys.
+    EMPTY_KEYS_SEGMENT = ["", "", "", ""].freeze
+
     def initialize(digests_key = DIGESTS)
       super(digests_key)
     end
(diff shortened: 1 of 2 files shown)


lib/sidekiq_unique_jobs/digests.rb
```diff
@@ -41,17 +41,27 @@ def delete_by_pattern(pattern, count: DEFAULT_COUNT)
     #   Also deletes the :AVAILABLE, :EXPIRED etc keys
     #
     # @param [String] digest a unique digest to delete
-    def delete_by_digest(digest) # rubocop:disable Metrics/MethodLength
+    # @param queuetime [Boolean] Whether to delete queue locks.
+    # @param runtime [Boolean] Whether to delete run locks.
+    def delete_by_digest(digest, queuetime: true, runtime: true) # rubocop:disable Metrics/MethodLength
       result, elapsed = timed do
+        queue = queuetime ? digest : ""
+        queue_queued = queuetime ? "#{digest}:QUEUED" : ""
```
mhenrixon: I'm still thinking about this, I get the problem, but wondering if there is a better way. 

The code ended up being a little repetitive below. Perhaps we could just have two other methods:

1. queuetime_keys
2. runtime_keys

```ruby
keys = []
keys.concat(queuetime_keys)
keys.concat(runtime_keys)

call_script(:delete_by_digest, keys)

EMPTY_KEYS_SEGMENT = ["", "", "", ""].freeze

def queuetime_keys(queuetime)
  return EMPTY_KEYS_SEGMENT unless queuetime
  
  [
    digest,
    "#{digest}:QUEUED",
    "#{digest}:PRIMED",
    "#{digest}:LOCKED",  
  ]
end
```

What do you think? I find the code a little harsh on the eyes. Do you know any better way (than mine) to construct the array to send to the script (the keys) I'm all ears and eyes.

bigzed: I refactored the code a little and created the two helper methods. WDYT?

mhenrixon (approved):
This is awesome. If you ensure that Reek passes, I'll release it!

Issue #782, opened 2023-05-19
`unlocked` reflection is never called


ttstarck:
**Describe the bug**
According to the [README](https://github.com/mhenrixon/sidekiq-unique-jobs#unlocked), there should be an `unlocked` reflection that is called when the lock is unlocked. However it appears that there isn't one at the moment. The closest thing I could find is a `reflect(:debug, [:unlocked...])` call in `Locksmith#unlock!`.

**Expected behavior**
When the lock is unlocked by the `Locksmith`, the `:unlocked` reflection is called.

**Current behavior**
When the lock is unlocked by the `Locksmith`, the `:debug` reflection is called with `:unlocked` as an argument.

I think this would be a simple fix, to add the `:unlocked` reflect right after the debug reflect. That way, this would not break anyone who may already have setup their reflects to be based off the `:debug`.

**Additional context**
[Pull Request](https://github.com/mhenrixon/sidekiq-unique-jobs/pull/783)

PR #783, opened 2023-05-19
Add unlocked reflect to locksmith unlock!

ttstarck:
https://github.com/mhenrixon/sidekiq-unique-jobs/issues/782


mhenrixon:
Thank you for the contribution @ttstarck!

Issue #785, opened 2023-05-26
Logging::Middleware.context can include empty string as key when lock is not set in job hash


ttstarck:
**Describe the bug**
Logs produced within the Middleware Logging context can contain an empty string as a key for jobs without `"lock"` key set in the job_hash. `""` as a key within the log context at the moment is causing these logs to not be able to be ingested into our logging backend (Elasticsearch).

**Expected behavior**
Logs produced within the Middleware Logging context properly sets the lock type used within the log context.

**Current behavior**
Logs produced within the Middleware Logging context includes `"" => digest` but doesn't include any context on what lock is being used.

**Additional context**
The [unique_disabled?](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/middleware.rb#L36) check will return true, even if `item[LOCK]` is false, if the job class' sidekiq_options contain a `lock`.

However, the middleware logging context, [called right after](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/main/lib/sidekiq_unique_jobs/middleware.rb#L40), assumes that `item[LOCK]` is set, and uses it as the key within the [logging_context method](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/main/lib/sidekiq_unique_jobs/logging/middleware_context.rb#L34-L37)

Two possible solutions for this bug:
1. Don't include the lock type within the log context if `item[LOCK]` is nil.
2. Add the lock to the job hash when we `prepare` the job.

My vote would be to go with Solution 2 as I could forsee other bugs coming up due to an assumption that `item[LOCK]` exists.

[PR for Solution 2](https://github.com/mhenrixon/sidekiq-unique-jobs/pull/786), but happy to hear other possible solutions.

ttstarck:
[Fix is merged](https://github.com/mhenrixon/sidekiq-unique-jobs/pull/786)

PR #786, opened 2023-05-26
Add lock to item when lock_type is pulled from Job class sidekiq options

ttstarck:
https://github.com/mhenrixon/sidekiq-unique-jobs/issues/785


mhenrixon (approved):
Brilliant! Thank you for your contribution.

Issue #779, opened 2023-05-05
Problems with reaper and long running jobs


RobsonKing:
**Describe the bug**
We have jobs that occasionally take a long time to execute (~20 minutes).  These jobs are using until_and_while_executing.  After about a minute a duplicate job start running.  The lock appears to be removed by the reaper (our reaper interval is 1 minute).

**Expected behavior**
Uniqueness should be maintained, even if the job runs for a long time.

**Current behavior**
- First job starts executing 
- Second job is enqueued,  gets blocked and put on the scheduled queue to retry.  So far everything is working as expected
```
TestingUniqueJob jid=e888a4c9373cceef87a7e834 INFO: start
TestingUniqueJob JID-e888a4c9373cceef87a7e834]With args [31] 
TestingUniqueJob JID-e888a4c9373cceef87a7e834 Enqueued TestingUniqueJob with [31] on default (85c4ea4da6a2f6df530a4178)
TestingUniqueJob jid=e888a4c9373cceef87a7e834 elapsed=0.057 INFO: done
```
- reaper runs and deletes the lock
```
uniquejobs=reaper INFO: Nothing to delete; exiting.
uniquejobs=reaper INFO: Deleting batch with 1 digests
INFO: (2023-05-04 21:57:00 +0000) Execution successfully returned 1
```
- job that was being blocked and rescheduled now starts executing.

**Worker class**

```ruby
class TestingUniqueJob <ApplicationSidekiqWorker
  sidekiq_options queue: 'default',
                  lock: :until_and_while_executing,
                  on_conflict: { client: :log, server: :reschedule }

  def perform(id)
    Rails.logger.info("*** [TestingUniqueJob] processing #{id} START")
    sleep(40 * 60)
    Rails.logger.info("*** [TestingUniqueJob] processing #{id} END")
  end
end
```

**Questions**
- Is this behaviour expected?
- Is there a way to prevent the reaper from deleting locks when jobs are still executing?
- How does the reaper determine that a lock can be cleaned up?

mhenrixon:
First of all, long running jobs are like ddosing yourself. 

https://twitter.com/nateberkopec/status/1635280301709746179?s=46

https://www.linkedin.com/posts/nate-berkopec-6832b41a8_long-running-background-jobs-60-seconds-activity-7037074319537016832-fite?utm_source=share&utm_medium=member_ios

That said, I'm sure there is something that could be done to improve. 

[...]

PR #789, opened 2023-06-07
Backport #756 to v7.1

andrepiske:
**Note**: I opened this branch from the `v7.1.29` tag, but I'm trying to merge it into the `v7.1` branch, which seems outdated (last commit Oct 2021 vs Dec 2022 for the tag). *My change is only 1 commit long*. I guess the v7.1 branch has to be updated with the v7.1.29 tag, as I can't open a PR to a tag. Or perhaps I'm doing something wrong here :)

I'm having exactly the same problem as https://github.com/mhenrixon/sidekiq-unique-jobs/issues/779 and this PR solves it by backporting PR https://github.com/mhenrixon/sidekiq-unique-jobs/pull/756, which applies to v8 of this gem, to v7.

I couldn't find any other instances of the string `:workers` that should be changed and I also tested this on my dev machine with a real app and it fixes the problem.

Let me know if this looks good enough!


sodabrew:
@andrepiske I'm researching something similar. This PR appears to include way too many commits. If you're comfortable with `git rebase` you can remove everything from this branch and instead run a `git cherry-pick` on the #756 commit so that you have only these 4 file changes: https://github.com/mhenrixon/sidekiq-unique-jobs/pull/756/files

mhenrixon:
This should be released as v7.1.30 https://github.com/mhenrixon/sidekiq-unique-jobs/releases/tag/v7.1.30

PR #805, opened 2023-08-17
Test improvements

Earlopain:
I don't run the test locally but in a docker compose setup, the hardcoded uris don't work well with that. Allow overriding them through env variables but fall back to the localhost ones.

Also make CI green by fixing rubocop violations.


mhenrixon (approved):

Earlopain:
Wow, that was fast (:

Issue #800, opened 2023-08-01
`while_executing` with `on_conflict: :reschedule` Reschedule job after job execution 


snyt45:
**Describe the bug**

I am not certain if this is a bug.
It started happening when I updated sidekiq unique-jobs from v8.0.2 to v8.0.3.

Environment in which it occurs：
- sidekiq v7.0.6
- sidekiq-unique-jobs v8.0.3

https://github.com/mhenrixon/sidekiq-unique-jobs/blob/f381562ba7a57748132ca3e988172f3021997d7c/lib/sidekiq_unique_jobs/lock/while_executing.rb#L41-L54

Events:
1. execute a Job whose lock strategy is `lock: :while_executing` and whose conflict strategy is `on_conflict: :reschedule`.
2. the Job is executed, but the contents of `executed` is empty and enters the condition of `unless executed` and the Job is rescheduled.

The above events loop infinitely.

**Expected behavior**

In v8.0.2, the job_id is placed in `executed` and is not rescheduled.

v8.0.2
```ruby
# lib/sidekiq_unique_jobs/lock/while_executing.rb
# Debugging code is embedded without changing the behavior of the original code itself.

      def execute(&block)
        with_logging_context do
          executed = locksmith.execute do
            yield
            if locksmith.unlock
              test_result1 = callback_safely
              puts "=== test_result1: #{test_result1}"
              test_result1
            end
          ensure
            test_result2 = locksmith.unlock
            puts "=== test_result2: #{test_result2}"
            test_result2
          end
          
          puts "=== executed: #{executed}"
          
          unless executed
            reflect(:execution_failed, item)
            call_strategy(origin: :server, &block)
          end
        end
      end

# 【result】
# === test_result1: f3a463a0f7d2256d93af271f
# === test_result2: false
# === executed: f3a463a0f7d2256d93af271f
```

**Current behavior**

In v8.0.3, the job is rescheduled because the `executed` is empty.

v8.0.3
```ruby
# lib/sidekiq_unique_jobs/lock/while_executing.rb
# Debugging code is embedded without changing the behavior of the original code itself.

      def execute(&block)
        with_logging_context do
          executed = locksmith.execute do
            test_result1 = yield
            puts "=== test_result1: #{test_result1}"
            test_result1  
          ensure
            test_result2 = unlock_and_callback
            puts "=== test_result2: #{test_result2}"
            test_result2
          end

          puts "=== executed: #{executed}"

          unless executed
[...]

PR #806, opened 2023-08-19
fix: `while_executing` should not invoke conflict strategy when the job was successfully executed.

leandrogoe:
Currently, the `WhileExecuting` lock only works properly if the middleware chain downstream returns a truthy value.

When a falsey value is returned, the on-conflict strategy is invoked, which shouldn't happen.

Originally reported [here](https://github.com/mhenrixon/sidekiq-unique-jobs/issues/800).

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/while_executing.rb
+++ b/lib/sidekiq_unique_jobs/lock/while_executing.rb
@@ -42,6 +42,7 @@ module SidekiqUniqueJobs
         with_logging_context do
           executed = locksmith.execute do
             yield
+            item[JID]
           ensure
             unlock_and_callback
           end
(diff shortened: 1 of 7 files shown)


mhenrixon (approved):

leandrogoe:
There were some linting failures unrelated with the files I've touched, not sure why. I've corrected those as well now to make sure the pipeline passes.

sfavrinlumint:
Hi @mhenrixon and @leandrogoe , was this fix supposed to go to the `v7.1` branch? Because the bug is present in the current release `8.0.3` and this PR fixes it.

leandrogoe:
@sfavrinlumint I made it for the 7.1 because I was using 7.1.30 and hence needed a fix in that branch. Another PR will be needed for 8.X

sfavrinlumint:
> @sfavrinlumint I made it for the 7.1 because I was using 7.1.30 and hence needed a fix in that branch. Another PR will be needed for 8.X

Ah silly me, gotcha thank you Leandro! I might open a PR for 8.0.3

RobsonKing:
Are there plans to publish a release for this fix?

PR #810, opened 2023-09-21
fix: `while_executing` should not invoke conflict strategy when the job was successfully executed [v8]

cuzik:
For context, I didn't find any PR fixing it to the 8 version, we are not updating the gem to v8.0.3 because of this error. I found the solution for the v7.1 here https://github.com/mhenrixon/sidekiq-unique-jobs/pull/806 and only replicated it. So if it works, all the credits go to @leandrogoe.

Let me know if it is something useful or if I missed something, but as far as I can see, we didn't open a PR fixing it to v8+, right?

It should fix the https://github.com/mhenrixon/sidekiq-unique-jobs/issues/800

### DIFF
--- a/lib/sidekiq_unique_jobs/lock/while_executing.rb
+++ b/lib/sidekiq_unique_jobs/lock/while_executing.rb
@@ -42,6 +42,7 @@ module SidekiqUniqueJobs
         with_logging_context do
           executed = locksmith.execute do
             yield
+            item[JID]
           ensure
             unlock_and_callback
           end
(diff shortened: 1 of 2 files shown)


cuzik:
The rubocop setting changed recently? Or was any dependency updated?

I tried to fix the error on the rubocop spec, but it found 59 new offenses on my branch. I don't think it is related to this PR, and as it is a lot of offenses, it would pollute the PR with the fixes.

Do you want me to open another PR to fix it, or I'm missing something?

mhenrixon (approved):

PR #813, opened 2023-10-16
Add sidekiq_option for on-conflict-reschedule perform_in time

mgmarlow:
Add schedule_in sidekiq job option to customize the on-conflict-reschedule duration of 5 seconds. The default remains 5 seconds if no schedule_in is provided.

In my particular use-case for on-conflict-reschedule, the jobs that I'm rescheduling have rather long processing times (eg minutes instead of seconds). The default `perform_in` of 5 seconds causes a lot of churn in the sidekiq scheduler, I'd prefer to schedule conflicts out another minute or so rather than attempting after a few seconds.

### DIFF
--- a/lib/sidekiq_unique_jobs/on_conflict/reschedule.rb
+++ b/lib/sidekiq_unique_jobs/on_conflict/reschedule.rb
@@ -21,7 +21,7 @@ module SidekiqUniqueJobs
       #   This will mess up sidekiq stats because a new job is created
       def call
         if sidekiq_job_class?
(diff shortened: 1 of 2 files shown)


mhenrixon:
Thank you for contributing! That's a great addition to the gem

mgmarlow:
> Thank you for contributing! That's a great addition to the gem

Happy to help!

I was taking a look at the build failures but I don't think they're related to this PR, unless I'm reading something wrong:

```
SidekiqUniqueJobs::Cli#help #help list displays help about the key command
```

Let me know if you'd like me to make any updates to this PR.

mhenrixon (approved):

PR #815, opened 2023-11-01
fix: Sidekiq 7.2 throws TypeError: Unsupported command argument type: TrueClass when using byscore: true 

Amnesthesia:
According to [Sidekiq 7.2 changelog](https://github.com/sidekiq/sidekiq/blob/main/Changes.md#720), using `byscore: true` and `nx: true` will now throw, which causes Sidekiq not to start up anymore when using `sidekiq-unique-jobs`

### DIFF
--- a/lib/sidekiq_unique_jobs/orphans/manager.rb
+++ b/lib/sidekiq_unique_jobs/orphans/manager.rb
@@ -192,7 +192,7 @@ module SidekiqUniqueJobs
       # @return [void]
       #
       def register_reaper_process
-        redis { |conn| conn.set(UNIQUE_REAPER, current_timestamp, nx: true, ex: drift_reaper_interval) }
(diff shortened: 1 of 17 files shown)


kwent:
Much needed ! Can we cut a new release with this patch ?

Amnesthesia:
> Much needed ! Can we cut a new release with this patch ?

Hope so, fixed linting and specs but workflow requires approval — hoping it's green now 🤞

Amnesthesia:
Any way you can approve workflow runs to run for the rest of the PR rather than for each commit? Doesn't it defeat the purpose of turning on auto-merge, if merge depends on workflows and workflows wont run without approval on each push?

mhenrixon:
> Any way you can approve workflow runs to run for the rest of the PR rather than for each commit? Doesn't it defeat the purpose of turning on auto-merge, if merge depends on workflows and workflows wont run without approval on each push?

I don't even care to approve. I need to check the settings. This is absolutely ridiculous. 

Also, I will remove reek. It seems a little overkill to be honest.

PR #827, opened 2024-01-29
Fix Testing Instructions

jherdman:
I ran into a known issue[1] when trying to bring this gem into my project.

It turns out the instructions were just out of date. If you merely include the RSpec file it misses other necessary requires. The noted file does everything you need regardless of testing framework.


mhenrixon (approved):

mhenrixon:
Thank you!

PR #829, opened 2024-02-05
fix(xss): sanitize parameters

mhenrixon:


Earlopain:
For anyone stumbling across this, recent versions of sidekiq already protect against this (at least the trivial cases I could quickly come up with myself). The CSP is more restrictive since version [7.2.0](https://github.com/sidekiq/sidekiq/blob/main/Changes.md#720), https://github.com/sidekiq/sidekiq/pull/6074, released October 31, 2023.

@mhenrixon would you mind creating/publishing a CVE so that the ecosystem can be properly be notified of this issue? This issue from sidekiq itself should serve as a pretty good base I think https://github.com/advisories/GHSA-h3r8-h5qw-4r35, basically the same thing.

mhenrixon:
Now I remember, I saw someone like to the correct form from their repository. 

I'll do the same in the future.

renchap:
Could you consider backporting this to version 7, for the people that are still using Sidekiq 6?

Thanks :)

mhenrixon:
> Could you consider backporting this to version 7, for the people that are still using Sidekiq 6?
> 
> Thanks :)

Are you not protecting your sidekiq routes with some type of admin constraint? 🤣 I'll get it one as soon as possible.

pboling:
I haven't seen a CVE for this yet, is one planned? @mhenrixon

mhenrixon:
Not yet, got busy with a bunch of other stuff.

I figured that I could piggy the sidekiq one.

pboling:
If I'm understanding this correctly, we have already upgraded past the fix for sidekiq, as we are currently on sidekiq v7.1.6, and IIRC it was fixed in sidekiq v7.0.8.  But we were one an old version of sidekiq-unique-jobs, and thus still had the vulnerability listed in this issue, I presume.

Does that make any sense?  Is a fix for this gem not needed if we have already upgraded sidekiq?  If not, then this gem needs its own separate CVE.

Earlopain:
If you are on `sidekiq` below version 7.2.0 and don't have a version of `sidekiq-unique-jobs` with this fix then you are vulnerable.

The change in 7.0.8 was for a similar XSS vulnerability like here, a fix in `sidekiq` doesn't fix it in `sidekiq-unique-jobs`. The change in `sidekiq` 7.2.0 was a preventive measure for similar problems, not in response to an actual issue and wasn't backported.

I agree that a CVE here would be necessary. It's not a problem if you are on latest `sidekiq` but like someone already mentioned they are on sidekiq 6.X which wasn't vulnerable to the original issue in sidekiq (and as such doesn't have the defensive fix) but does have the problem this PR fixed (as such the desire for a backport of this).

Even though everyone should have the admin dashboard behind a constraint, knowledge that the dashboard exists and getting a user to click a malicious link is good enough for exploitation.

pboling:
Screenshot 2024-02-12 at 16 29 38
I tried.  Looks like @mhenrixon is the hero we need, unless another collaborator / maintainer has access.

mhenrixon:
> This is an important note. Social hacking, where an attacker gets an authorized users to click on a malicious link, is often the easiest kind.

Looks like I had to enable advisories for it to work. It is enabled now

pboling:
Thanks!  Reporting...

lib/sidekiq_unique_jobs/web.rb
```diff
@@ -71,20 +71,20 @@ def self.registered(app) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
...
+        digests.delete_by_digest(h(params[:digest]))
+        expiring_digests.delete_by_digest(h(params[:digest]))
         redirect_to :locks
       end
 
       app.get "/locks/:digest/jobs/:job_id/delete" do
-        @digest = params[:digest]
+        @digest = h(params[:digest])
         @lock   = SidekiqUniqueJobs::Lock.new(@digest)
         @lock.unlock(params[:job_id])
```
pboling, line 89: @mhenrixon 
Both `params[:digest]` and `params[:job_id]` are derived from the request path, but only `params[:digest]` got the fix.  Is there a simple reason?

Earlopain: This parameter is not displayed in the web UI, you're free to unlock by `` if you like but its not going to do much. In a similar vein the changes to `/locks/:digest/delete` seem unnecessary.

pboling: Thanks @Earlopain, so the fix is good, if a little overzealous.

When some are, and some are not, escaped, it might be good to document the reasons @mhenrixon, so people don't have to wonder, or become alarmed that there may be latent XSS issues here, when there are not any.

pboling: To be clear then, only three endpoints were vulnerable *prior to this change*, and the changes here included some over protective escaping on some routes.

The vulnerable routes in the CVE are:

1. `/changelogs`
2. `/locks`
3. `/expiring_locks`

mhenrixon: Ah, must have opened the wrong view. I just looked at the usage in the view, as the internal usage on the ruby side isn't affected. 

Searched for @digest and escaped the parameter.

pboling: That's great for consistency, and no one is load testing the admin UI, so the perf hit is immaterial, even if escaping the path-derived params isn't required.  Would it make sense to also escape the `job_id` for consistency though?

pboling: I'm almost done reporting the CVE!

mhenrixon: > ould it make sense to also escape the `job_id` for consistency though?

Absolutely, the idea was to escape everything but I honestly didn't see the job_id params.

I'll get that sorted and create a backport later today. I have to take my dog to the vet first.

pboling: @mhenrixon If you have time for a quick call it would be good. :) I went deep on this.

mhenrixon: Sure thing @pboling, I just have to perform some self-care first. I should be available in an hour or so. 

Preferred means of communication?

pboling: @mhenrixon I'll be in my zoom room for the next hour and a half.  Emailing you the link.

pboling:
@Earlopain new CVE requested for XSS & RCE.  Thanks for your help identifying!

mhenrixon:
Massive thank you to both @Earlopain and @pboling.

The CWE is now not only published but also fixed. 

Couldn't have done it without you guys! The repository will be more secure moving forward.

PR #833, opened 2024-02-12
fix(rce): prevent remot code execution

mhenrixon:


Earlopain:
Sorry for the bother again. Is this what you were talking about in https://github.com/mhenrixon/sidekiq-unique-jobs/pull/829#discussion_r1486014158? I'd assume so but the title/commit makes this sound way worse.

mhenrixon:
> Sorry for the bother again. Is this what you were talking about in https://github.com/mhenrixon/sidekiq-unique-jobs/pull/829#discussion_r1486014158? I'd assume so but the title/commit makes this sound way worse.
> 
> 
> 
> 8.0.10 isn't pushed to rubygems yet.

It was way worse, redis doesn't filter anything according to @pboling who kindly provided a way to replicate the problem.

8.0.10 is unreleased because it contains no changes. 

The fix was released as 8.0.9

Earlopain:
I see, thanks for the info. My bad on reading the commit history wrong, I see now that it is like you said.

Great work from you and @pboling for identifying and fixing this problem.

pboling:
> I see, thanks for the info. My bad on reading the commit history wrong, I see now that it is like you said.
> 
> Great work from you and @pboling for identifying and fixing this problem.

I couldn't have done it without your multiple explanations of the context and the relevant issues.  I'll submit you for credit on the CVE request, which you can accept if you like.

Earlopain:
I'm not certain if this is actually an issue. Take these commands for example:
```irb
irb(main):002> Cache.redis.hset("hash", "]\"key\"", "value")
=> 1
irb(main):003> Cache.redis.hgetall("hash")
=> {"]\"key\""=>"value"}
irb(main):004> Cache.redis.hdel("hash", "]\"key\"")
=> 1
irb(main):005> Cache.redis.hgetall("hash")
=> {}
```

The key contains `]` as well as `"` literals but redis/redis-client is taking them literally. https://redis.io/docs/management/security/ says:
> *String escaping and NoSQL injection*
> The Redis protocol has no concept of string escaping, so injection is impossible under normal circumstances using a normal client library. The protocol uses prefixed-length strings and is completely binary safe.

In addition, using the `h` function to facilitate escaping seems brittle. Its purpose has nothing to do with redis so anything it solves is by chance at best. There may be other special characters that have meaning in redis but are not relevant for escaping untrusted input for browser consumption.

I myself didn't manage to reproduce but if I did something wrong then I believe this should 100% be reported upstream to `redis-client`.

pboling:
I see what you mean. Looking at this more soberly - It is the argument to the `hdel` command that is being injected from the path parameter here.  The following queries that I hastily mistook as the result of an INFO command running were caused by the redirect back to the sidekiq `/locks` page gathering data for rendering itself.

RCE - redis server

I'll update the advisory.

pboling:
This does still mean that a hacker could construct a URL that would result in deletion of targeted keys, but also there is no fix for that.

Earlopain:
That's a relief, a 10/10 is quite severe (by definition). Thanks for double-checking.

I'm not certain what the layout of that hash is exactly, or if the key is dynamic in some way. I would expect it to only contain information about the status of locks, so the most you can do there is what you can already do through the web UI.

Regardless, you can do a bunch of stuff with the Web UI with plain sidekiq already like deleting complete queues so I don't think this should be a big deal. Unless you can delete arbitrary keys, that would be a bit of a problem. From a quick peek at the code that doesn't look possible to introduce arbitrary digests.

pboling:
I think it is still a 10/10, since the XSS can expose cookies, session and local storage.

I've corrected the advisory.  Please give any feedback you have!

Earlopain:
That looks better but it's definitely not a 10. The sidekiq equivalent is rated 8.3 which I think is still quite high but eh. I would suggest just copying most of the values from https://github.com/advisories/GHSA-h3r8-h5qw-4r35

While XSS is bad, it's not RCE levels kind of bad.

In general these reports with their severities fail to take into account that you are expected to put the dashboard behind a constraint. That's a separate issue though and not something I want to debate. While I don't personally agree, it's _fine, I guess.

Earlopain:
I'm not an expert on this, I haven't read the actual definitions. I just find it laughable that the difference between RCE and XSS is 0.6 points. But that has nothing to do with you, if that is indeed how this works then I'm just shaking my head in defeat at the process in general.

Anyways, I believe @mhenrixon still needs to [request a CVE](https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/publishing-a-repository-security-advisory#requesting-a-cve-identification-number-optional) for this to actually be picked up by tools like dependabot.

I believe the report should be compacted a bit beforehand. I don't think I've seen such details before, including POC.
Take these as an example:
* https://github.com/advisories/GHSA-68xg-gqqm-vgj8
* https://github.com/advisories/GHSA-3hhc-qp5v-9p2j
* https://github.com/advisories/GHSA-wq4h-7r42-5hrr
* https://github.com/advisories/GHSA-hq7p-j377-6v63

They don't go into nearly as much detail. I would find it fine as initial disclosure but not something consumers should see in the end. I would instead instead link to the relevant PRs as reference where interested users can get more information. Particularly the commit with the fix should be there.

mhenrixon:
Appreciate the conversation, I did request the CVE already but unsure how it works after. 

Everywhere I look is simply overwhelming information. Find it hard to narrow things down. 

Next time we can perhaps be a bit briefer.

Earlopain:
Eh, perhaps they'll trim it down a bit, or it updates transparently. What's done is done. From what I read, just waiting now should do the job.

Unfortunatly there's no indicator that you've already done so, my bad. At least on the GitHub advisory side these things do update, they are simply stored in git and you can even do PRs against that.

pboling:
It says the CVE process can take a few days, and since it was just requested a few hours ago, we may have a few days wait.

PR #835, opened 2024-02-13
Add digest scores for faster deletes in sorted sets

### LINKED ISSUE #668: Slow evalsha causing timeouts
Hi Mikael, first of all thank you so much for your work on this lib ! We are very satisfied with it. I've tried my best below to be as precise and provide as much context as possible, as I believe the issue is related to sidekiq-unique-jobs, but I am aware I could be mistaking.

**Describe the bug**
Our workers are reporting a lot of `Redis::TimeoutError` `connection timed out` errors. On the server I can see redis is using 100% cpu (one core). The slowlog is filled with entries that I can to some degree trace back to `sidekiq-unique-jobs` through the `delete_job_by_digest` argument: see 1) 4) 11). The entries look like this:

```
1) (integer) 226683
   2) (integer) 1638716967
   3) (integer) 935819
   4)  1) "evalsha"
       2) "14ab482616f7d7df8b71783a1c3db6eb2103a2a6"
       3) "3"
       4) "queue:low"
       5) "schedule"
       6) "retry"
       7) "uniquejobs:3ee9129f6972ac426c645a6b1b4b845e"
       8) "1638716966.1943657"
       9) "false"
      10) "1000"
      11) "delete_job_by_digest"
      12) "5.0.7"
   5) "172.17.0.14:55328"
   6) ""
```
When the queues are small, this does not seem to be an issue, but when we increased from 180.000 to 500.000 enqueued jobs, the amount of timeouts increase tremendously. We used to have 60 concurrent workers of this class without any issue. What stands out for me: we have not defined any unique-jobs specific config in the worker, and the worker itself does nothing complex.

**Expected behavior**
No timeouts 

**Current behavior**
±150 very slow `evalsha` calls

**Worker class**

```ruby
class MyDearWorker
  include Sidekiq::Worker
  include Sidekiq::Throttled::Worker

  sidekiq_options(
    queue: :low,
    retry: 8
  )

  # sidekiq_throttle config

  def perform
    # do something
  end
end
```

**Additional context**

SidekiqUniqueJobs was updated today from `7.0.8` to `7.1.12` to no avail
Sidekiq `6.3.1`

`redis-cli info` returns
```
# Server
redis_version:5.0.7
redis_git_sha1:00000000
redis_git_dirty:0
redis_build_id:5f6bfe2b13cc4617
redis_mode:standalone
os:Linux 4.15.0-74-generic x86_64
arch_bits:64
multiplexing_api:epoll
atomicvar_api:atomic-builtin
gcc_version:8.3.0
process_id:1
run_id:1afc9ece9143d52e06cc7dcd1e2ea967779735e2
tcp_port:6379
uptime_in_seconds:2744770
uptime_in_days:31
hz:10
configured_hz:10
lru_clock:11328868
executable:/data/redis-server
config_file:/usr/local/etc/redis/redis.conf

# Clients
connected_clients:79
client_recent_max_input_buffer:2
client_recent_max_output_buffer:0
blocked_clients:64

# Memory
used_memory:349476912
used_memory_human:333.29M
used_memory_rss:360284160
used_memory_rss_human:343.59M
used_memory_peak:367916168
used_memory_peak_human:350.87M
used_memory_peak_perc:94.99%
used_memory_overhead:3616774
used_memory_startup:791504
used_memory_dataset:345860138
used_memory_dataset_perc:99.19%
allocator_allocated:349562064
allocator_active:350646272
allocator_resident:361467904
total_system_memory:67543748608
total_system_memory_human:62.91G
used_memory_lua:103424
used_memory_lua_human:101.00K
used_memory_scripts:35744
used_memory_scripts_human:34.91K
number_of_cached_scripts:12
maxmemory:0
maxmemory_human:0B
maxmemory_policy:noeviction
allocator_frag_ratio:1.00
allocator_frag_bytes:1084208
allocator_rss_ratio:1.03
allocator_rss_bytes:10821632
rss_overhead_ratio:1.00
rss_overhead_bytes:-1183744
mem_fragmentation_ratio:1.03
mem_fragmentation_bytes:10767344
mem_not_counted_for_evict:0
mem_replication_backlog:0
mem_clients_slaves:0
mem_clients_normal:2647718
mem_aof_buffer:0
mem_allocator:jemalloc-5.1.0
active_defrag_running:0
lazyfree_pending_objects:0

# Persistence
loading:0
rdb_changes_since_last_save:863
rdb_bgsave_in_progress:0
rdb_last_save_time:1638718611
rdb_last_bgsave_status:ok
rdb_last_bgsave_time_sec:0
rdb_current_bgsave_time_sec:-1
rdb_last_cow_size:2232320
aof_enabled:0
aof_rewrite_in_progress:0
aof_rewrite_scheduled:0
aof_last_rewrite_time_sec:-1
aof_current_rewrite_time_sec:-1
aof_last_bgrewrite_status:ok
aof_last_write_status:ok
aof_last_cow_size:0

# Stats
total_connections_received:221466
total_commands_processed:1636072405
instantaneous_ops_per_sec:72
total_net_input_bytes:93096542534
total_net_output_bytes:53079207633
instantaneous_input_kbps:19.85
instantaneous_output_kbps:7.00
rejected_connections:0
sync_full:0
sync_partial_ok:0
sync_partial_err:0
expired_keys:47100
expired_stale_perc:0.14
expired_time_cap_reached_count:0
evicted_keys:0
keyspace_hits:1508620767
keyspace_misses:2491703
pubsub_channels:1
pubsub_patterns:0
latest_fork_usec:10812
migrate_cached_sockets:0
slave_expires_tracked_keys:0
active_defrag_hits:0
active_defrag_misses:0
active_defrag_key_hits:0
active_defrag_key_misses:0

# Replication
role:master
connected_slaves:0
master_replid:18bafaa2c497072bb991b4bdade7721b38155c46
master_replid2:0000000000000000000000000000000000000000
master_repl_offset:0
second_repl_offset:-1
repl_backlog_active:0
repl_backlog_size:1048576
repl_backlog_first_byte_offset:0
repl_backlog_histlen:0

# CPU
used_cpu_sys:3475.941359
used_cpu_user:134248.522011
used_cpu_sys_children:67.156182
used_cpu_user_children:596.732377

# Cluster
cluster_enabled:0

# Keyspace
db0:keys=1507,expires=1349,avg_ttl=127866632386
```

Does this ring any bells? Thanks a lot in advance, if I can provide any more info please let me know

ezekg:
Closes #668. First pass on [my idea](https://github.com/mhenrixon/sidekiq-unique-jobs/issues/668#issuecomment-1934461714) of using scores to "skip ahead" so that we don't have to iterate the entire sorted set when a unique job is deleted. Let me know what you think. I'm sure I'm missing some edge cases, since I don't have the domain knowledge you do here. This _should_ be backwards compatible, reverting back to previous behavior if a score doesn't exist.

I still want to add a performance test with a large schedule queue to make sure this actually works.

### DIFF
--- a/lib/sidekiq_unique_jobs/locksmith.rb
+++ b/lib/sidekiq_unique_jobs/locksmith.rb
@@ -82,7 +82,7 @@ module SidekiqUniqueJobs
     # Deletes the lock regardless of if it has a pttl set
     #
     def delete!
-      call_script(:delete, key.to_a, [job_id, config.pttl, config.type, config.limit]).to_i.positive?
+      call_script(:delete, key.to_a, argv).to_i.positive?
     end
 
     #
@@ -362,7 +362,11 @@ module SidekiqUniqueJobs
     end
 
     def argv
-      [job_id, config.pttl, config.type, config.limit]
+      [job_id, config.pttl, config.type, config.limit, lock_score]
+    end
+
+    def lock_score
+      item[AT].to_s
     end
 
     def lock_info
@@ -375,6 +379,7 @@ module SidekiqUniqueJobs
         TYPE => config.type,
         LOCK_ARGS => item[LOCK_ARGS],
         TIME => now_f,
+        AT => item[AT],
       )
     end
 
--- a/lib/sidekiq_unique_jobs/lua/lock.lua
+++ b/lib/sidekiq_unique_jobs/lua/lock.lua
@@ -15,15 +15,16 @@ local job_id       = ARGV[1]
 local pttl         = tonumber(ARGV[2])
 local lock_type    = ARGV[3]
 local limit        = tonumber(ARGV[4])
+local lock_score   = ARGV[5]
 -------- END lock arguments -----------
 
 
 --------  BEGIN injected arguments --------
-local current_time = tonumber(ARGV[5])
-local debug_lua    = tostring(ARGV[6]) == "1"
-local max_history  = tonumber(ARGV[7])
-local script_name  = tostring(ARGV[8]) .. ".lua"
-local redisversion = ARGV[9]
+local current_time = tonumber(ARGV[6])
+local debug_lua    = tostring(ARGV[7]) == "1"
+local max_history  = tonumber(ARGV[8])
+local script_name  = tostring(ARGV[9]) .. ".lua"
+local redisversion = ARGV[10]
 ---------  END injected arguments ---------
 
 
@@ -62,8 +63,16 @@ if lock_type == "until_expired" and pttl and pttl > 0 then
   log_debug("ZADD", expiring_digests, current_time + pttl, digest)
   redis.call("ZADD", expiring_digests, current_time + pttl, digest)
 else
-  log_debug("ZADD", digests, current_time, digest)
-  redis.call("ZADD", digests, current_time, digest)
+  local score
+
+  if #lock_score == 0 then
+    score = current_time
+  else
+    score = lock_score
+  end
+
+  log_debug("ZADD", digests, score, digest)
+  redis.call("ZADD", digests, score, digest)
 end
 
 log_debug("HSET", locked, job_id, current_time)
--- a/lib/sidekiq_unique_jobs/lua/shared/_delete_from_sorted_set.lua
+++ b/lib/sidekiq_unique_jobs/lua/shared/_delete_from_sorted_set.lua
@@ -1,19 +1,29 @@
 local function delete_from_sorted_set(name, digest)
-  local per   = 50
-  local total = redis.call("zcard", name)
-  local index = 0
-  local result
+  local score  = redis.call("ZSCORE", "uniquejobs:digests", digest)
+  local total  = redis.call("ZCARD", name)
+  local per    = 50
 
-  while (index < total) do
-    local items = redis.call("ZRANGE", name, index, index + per -1)
+  for offset = 0, total, per do
+    local items
+
+    if score then
+      items = redis.call("ZRANGE", name, score, "+inf", "BYSCORE", "LIMIT", offset, per)
+    else
+      items = redis.call("ZRANGE", name, offset, offset + per -1)
+    end
+
+    if #items == 0 then
+      break
+    end
+
     for _, item in pairs(items) do
       if string.find(item, digest) then
         redis.call("ZREM", name, item)
-        result = item
-        break
+
+        return item
       end
     end
-    index = index + per
   end
-  return result
+
+  return nil
 end
--- a/spec/performance/unique_job_on_conflict_replace_spec.rb
+++ b/spec/performance/unique_job_on_conflict_replace_spec.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+RSpec.describe UniqueJobOnConflictReplace, :perf do
+  let(:lock_prefix)  { described_class.sidekiq_options.fetch("lock_prefix")  { SidekiqUniqueJobs.config.lock_prefix } }
+  let(:lock_timeout) { described_class.sidekiq_options.fetch("lock_timeout") { SidekiqUniqueJobs.config.lock_timeout } }
+  let(:lock_ttl)     { described_class.sidekiq_options.fetch("lock_ttl")     { SidekiqUniqueJobs.config.lock_ttl } }
+  let(:queue)        { described_class.sidekiq_options["queue"] }
+  let(:on_conflict)  { described_class.sidekiq_options["on_conflict"] }
+  let(:lock)         { described_class.sidekiq_options["lock"] }
+
+  before do
+    flushdb
+  end
+
+  context "when schedule queue is large" do
+    it "locks and replaces quickly" do
+      (0..100_000).each_slice(1_000) do |nums|
+        redis do |conn|
+          conn.pipelined do |pipeline|
+            nums.each do |num|
+              created_at   = Time.now.to_f
+              scheduled_at = created_at + rand(3_600..2_592_000)
+
+              payload = {
+                "retry" => true,
+                "queue" => queue,
+                "lock" => lock,
+                "on_conflict" => on_conflict,
+                "class" => described_class.name,
+                "args" => [num, { "type" => "extremely unique" }],
+                "jid" => SecureRandom.hex(12),
+                "created_at" => created_at,
+                "lock_timeout" => lock_timeout,
+                "lock_ttl" => lock_ttl,
+                "lock_prefix" => lock_prefix,
+                "lock_args" => [num, { "type" => "extremely unique" }],
+                "lock_digest" => "#{lock_prefix}:#{SecureRandom.hex}",
+              }
+
+              pipeline.zadd("schedule", scheduled_at, payload.to_json)
+            end
+          end
+        end
+      end
+
+      # queueing it once at the end of the queue should succeed
+      expect(described_class.perform_in(2_592_000, 100_000, { "type" => "extremely unique" })).not_to be_nil
+
+      # queueing it again should be performant
+      expect do
+        described_class.perform_in(2_592_000, 100_000, { "type" => "extremely unique" })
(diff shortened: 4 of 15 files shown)


lib/sidekiq_unique_jobs/locksmith.rb
```diff
@@ -378,6 +394,14 @@ def lock_info
       )
     end
 
+    def lock_score
+      @lock_score ||= item[CREATED_AT]
```
ezekg: This isn't right. It should be the time at which the job is scheduled, and I don't think this is that.

Trying to figure out where I can get that info, since it [seems transient](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/api.rb#L667).

lib/sidekiq_unique_jobs/locksmith.rb
```diff
@@ -331,6 +333,20 @@ def write_lock_info(conn)
...
+    #   The lock score is used to improve performance when iterating large sorted
+    #   sets e.g the schedule and retry sets.
+    #
+    #
+    # @return [void]
+    #
+    def write_lock_score(conn)
+      return unless lock_score?
+
+      conn.set(key.score, lock_score)
```
ezekg: Need to add tests for when this gets set, i.e. only for the `schedule` and `retry` queues.

lib/sidekiq_unique_jobs/key.rb
```diff
@@ -37,6 +37,10 @@ class Key
     # @!attribute [r] expiring_digests
     #   @return [String] the zset with locked expiring_digests
     attr_reader :expiring_digests
+    #
+    # @!attribute [r] score
+    #   @return [String] score (timestamp) for the lock
```
ezekg: s/timestamp/scheduled at timestamp/

mhenrixon:
Let's think about this for a while. We already store a score when the digest is created. It has both a score and a job_id. We could simplify this greatly by looking up the current digest's score and job id.

ezekg:
> We already store a score when the digest is created. It has both a score and a job_id.

Ah, I wasn't aware we were already storing the score. I looked around but I guess I missed it. That makes things easier.

> Sidekiq only checks with the heartbeat method every ten seconds. This means that for ten seconds, things can be missing.

What does Sidekiq's heartbeat system do? I'm not familiar with it and what "missing" means here.

ezekg:
Added a rudimentary performance test to showcase the issue and how this approach fixes it.

lib/sidekiq_unique_jobs/lua/shared/_delete_from_sorted_set.lua
```diff
@@ -1,19 +1,32 @@
...
 
-  while (index < total) do
-    local items = redis.call("ZRANGE", name, index, index + per -1)
     for _, item in pairs(items) do
       if string.find(item, digest) then
         redis.call("ZREM", name, item)
-        result = item
-        break
+
+        return item
```
ezekg, line 23: Is there a particular reason this didn't have a base case before? It would iterate the entire set even if a result was found, but that doesn't make sense at first glance because a unique job can only exist once in the set.

The `delete_from_queue` script also lacks a base case.

ezekg:
@mhenrixon I added a rudimentary performance test that fails for the old algorithm.

mhenrixon:
I really appreciate you taking a stab at this @ezekg! You spotted a couple of bugs nicely, first of all, unnecessary looping. Secondly, zrange is super slow in this case because it loops through everything. We should be using zscan, of course. 

Can you check if the linked PR fixes your problem, too? I prefer a simpler fix than adding these extra keys. If it still isn't good enough, I'm happy to pair on sorting this out!

spec/performance/unique_job_on_conflict_replace_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+            end
+          end
+        end
+      end
+
+      # queueing it once at the end of the queue should succeed
+      expect(described_class.perform_in(2_592_000, 100_000, { "type" => "extremely unique" })).to_not be_nil
+
+      # queueing it again should be performant
+      expect { Timeout.timeout(0.1) { described_class.perform_in(2_592_000, 100_000, { "type" => "extremely unique" }) } }.to_not raise_error
```
ezekg: Replace `Timeout` with `expect { ... }.to perform_under(10).ms`.

mhenrixon:
If you have a look here: https://github.com/mhenrixon/sidekiq-unique-jobs/blob/main/lib/sidekiq_unique_jobs/lua/lock.lua#L66 we already add a score for the digest, we also store the timestamp in he hash with digest + job_id (this is to be able to allow concurrent jobs of a specified number): https://github.com/mhenrixon/sidekiq-unique-jobs/blob/main/lib/sidekiq_unique_jobs/lua/lock.lua#L70

Between those two, it is a hard sell to add more timestamps. I am more inclined to attempt to reduce the number of commands than to increase them.

Could you find a way to make do with what is there or do a great job at selling the extra key to me @ezekg? I want to help you, I really do!

ezekg:
@mhenrixon the current implementation does not store the actual score of the job in the sorted set (i.e. the timestamp at which a scheduled job is scheduled to run at), but rather the current time at which the job is added to the sorted set. I updated the implementation to now store the job's score when available. I needed this to match the job's score in Sidekiq's `schedule` sorted set, otherwise it's useless in performing a divide-and-conquer search.

Let me know if you have time to review the new approach.

lib/sidekiq_unique_jobs/lua/lock.lua
```diff
@@ -62,8 +63,16 @@ if lock_type == "until_expired" and pttl and pttl > 0 then
   log_debug("ZADD", expiring_digests, current_time + pttl, digest)
   redis.call("ZADD", expiring_digests, current_time + pttl, digest)
```
ezekg, line 64: I left this alone because I don't fully understand what the expiring digests set is used for and if this optimization would be applicable to jobs utilizing an `until_expired` lock strategy.

spec/support/sidekiq_unique_jobs/testing.rb
```diff
@@ -59,12 +59,12 @@ def debug(*args)
...
 
-      def flushall(options = nil)
-        redis { |conn| conn.flushall(options) }
+      def flushall(...)
+        redis { |conn| conn.flushall(...) }
       end
 
-      def flushdb(options = nil)
-        redis { |conn| conn.flushdb(options) }
+      def flushdb(...)
```
ezekg, line 66: This and `flushall` were broken for the default case, i.e. where `nil` was provided.

mhenrixon (approved):
This is exactly what I had in mind. Sorry that I haven't been able to get to it. 

Your assumption about the job score in the hash is correct, I just wanted something in the hash and wasn't clear with how to use it. It was one of those: "I believe I will need this".

As soon as the German IRS is off my back (they want money I don't have), I'll see about greatly optimizing the gem. 

I never got around to looking at the performance.

Perhaps this would be better solved in ruby (like the reaper).

I'm definitely not opposed to use some batching from the ruby layer if there are more than n number in a sorted set. 

I believe there are plenty to optimize and for the performance tests I need to remember that my machine is as fast as a laptop comes. 

Not fair to compare locally, should probably write a bunch of these performance tests and have them run on GitHub actions.

ezekg:
Do you want me to loop in the changes from #837? That had improvements to `delete_from_queue` as well, which I didn't touch here. Let me know if you want me to cherry-pick, or if you want to do that separately.

mhenrixon:
> Do you want me to loop in the changes from #837?

That would be lovely, I forgot about the queue! I'm too exhausted to cut a release and test this tonight but I will do it first thing tomorrow morning. 

Much appreciated!!

ezekg:
Done. Let me know what you find whenever you have a chance to test.

mhenrixon:
Looks like it will do the job just fine! Can't wait to optimize everything!

PR #836, opened 2024-02-16
tweak changelog for 8.0.9

Earlopain:
Makes an explicit note that this was a non-issue. Dependabot uses this changelog file to provide context snippets in PR descriptions. My hope is that even though there has been no new release this will be picked up anyways.

Alternativly, the changelog entry could be removed entirely.

Would need backporting to 7.1 I think

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 **Fixed bugs:**
 
+- note: The RCE vulnerability was a false alarm, `sidekiq-unique-jobs` was not vulnerable to RCE. You can find additional information in the PR linked below.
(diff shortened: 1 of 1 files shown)


mhenrixon (approved):
The problem is that the changelog is generated based on git history. 

I should likely maintain it manually because I am unhappy with the result. 

Perhaps I generate a temporary one and maintain the real one manually. 

I appreciate you bringing this to my attention. 

I'll make sure to copy this to the release notes on the GitHub release as well.

mhenrixon:
I won't merge yet, want to prevent your addition from magically disappearing like my commit the other day.

mhenrixon:
Added it to the Github release for both gem versions.

Earlopain:
> won't merge yet

Merged 3 minutes later. Thanks. Again, not sure if tools will actually pick this up or if they pull from specific releases instead.

Fetching the latest one would make sense, to make amends like this. Let's see ehat happens.

PR #867, opened 2025-02-26
Remove redundant include to locales (for Sidekiq 8)

jukra:
This generates errors for Sidekiq 8, and looks like this gem does not have any translation files, so it can be just removed?
```
ERROR: undefined method 'settings' for class Sidekiq::Web
```


mhenrixon (approved):

jukra:
Thanks for the review! @mhenrixon could you  merge this? I think the failing ci for rubocop should be fixed in a separate PR, as it looks like to be about outdated config in general?

PR #868, opened 2025-02-28
update Reject#kill_with_options? for Ruby 3 kwargs

stathis-alexander:
Sometime in the past either the definition of `arity` or `Sidekiq::Deadset#kill` was changed. Currently https://apidock.com/ruby/Method/arity returns negative numbers when there are optional arguments, which this method does not consider. The current (and what's been the state for many years now) version of this method takes an optional hash / `opts` argument": https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/api.rb#L900, so the current method is just always returning `false`. 

My suspicion is that this changed around Ruby3 with the kwarg changes, but I'm not sure. 

This updates the method to return `true` if it takes more than one required argument, whether there are optional args or not.

### DIFF
--- a/lib/sidekiq_unique_jobs/on_conflict/reject.rb
+++ b/lib/sidekiq_unique_jobs/on_conflict/reject.rb
@@ -28,7 +28,8 @@ module SidekiqUniqueJobs
       # @return [false] when Sidekiq::Deadset#kill does not take multiple arguments
       #
       def kill_with_options?
-        Sidekiq::DeadSet.instance_method(:kill).arity > 1
+        kill_arity = Sidekiq::DeadSet.instance_method(:kill).arity
+        kill_arity > 1 || kill_arity < -1
       end
 
       #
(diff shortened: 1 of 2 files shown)


lib/sidekiq_unique_jobs/on_conflict/reject.rb
```diff
@@ -28,7 +28,8 @@ def call
       # @return [false] when Sidekiq::Deadset#kill does not take multiple arguments
       #
       def kill_with_options?
-        Sidekiq::DeadSet.instance_method(:kill).arity > 1
+        kill_arity = Sidekiq::DeadSet.instance_method(:kill).arity
+        kill_arity > 1 || kill_arity < -1
```
mhenrixon, line 40: Could you comment why the double check is here and what the two different variants mean specifically?

I am positive I will forget what this is all about!

Thank you for the contribution.

stathis-alexander: Done. Let me know if you'd like me to alter the comment in some way.

mhenrixon (approved):

Issue #882, opened 2025-06-06
Set ActiveSupport::Duration to lock_ttl, the key is not expired


tazoo:
**Describe the bug**

In my code I set `ActiveSupport::Duration` to `lock_ttl:`.
After upgrading to version 8.0.11, lock_tll is not working.
I think it's because of #879.
This PR doesn't take ActiveSupport::Duration into account.

**Expected behavior**
if `lock_ttl: 1.minute`, lock_ttl work as 60.seconds.
Or, if you set Duration, some error will raise.

**Current behavior**
if `lock_ttl: 1.minute`, lock_ttl work as `nil`.

**Worker class**

```ruby
class MyWorker
  include Sidekiq::Worker
  sidekiq_options lock: :until_executed, queue: :undefault, lock_ttl: 1.minute
  def perform(args); end
end
```

**Additional context**
The README describes a pattern for setting Duration in lock_ttl and a pattern for setting to_i.
This is very confusing, so please fix it.

https://github.com/mhenrixon/sidekiq-unique-jobs/blob/73386b1c4a3c4af00c7db1e1503e87f59f1771bd/README.md?plain=1#L242
https://github.com/mhenrixon/sidekiq-unique-jobs/blob/73386b1c4a3c4af00c7db1e1503e87f59f1771bd/README.md?plain=1#L906

PR #884, opened 2025-06-13
Raise unexpected type error for lock ttl

tazoo:
fix for https://github.com/mhenrixon/sidekiq-unique-jobs/issues/882

ArgumentError is now raised if lock_ttl is an unexpected type.
Also, the README has been modified accordingly.

This change is to prevent returning nil with an unexpected type and tying up the lock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ This lock behaves identically to the [Until Executed](#until-executed) except fo
 class UntilExpired
   include Sidekiq::Workers
 
-  sidekiq_options lock: :until_expired, lock_ttl: 1.day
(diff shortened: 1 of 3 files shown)


luis30n:
@mhenrixon sorry to bother. 
The lock_ttl integer enforcement was unintentionally introduced at https://github.com/mhenrixon/sidekiq-unique-jobs/pull/879 so if some codebases (like mine) have lock_ttl: Duration, lots of jobs would have acquired never expiring locks by now.
I'm preparing a script to clean them up for my company and I was wondering if you'd consider it worth of having a rake task to help others with this cleanup. 
Let me know what you think

Issue #850, opened 2024-07-18
lock_prefix argument is not respected


adamcreekroad:
**Describe the bug**

Currently, the `lock_prefix` configuration option is ignored, and all created locks currently all use the global default `uniquejobs`.

I believe this to be a regression from #774, specifically [this LOC](https://github.com/mhenrixon/sidekiq-unique-jobs/pull/774/files#diff-94f43e45beddea58a894d3689b611d2ee718a601f2bef966bae3fb57cc382b38R56). The way the code is now, it always sets the prefix to the global default, rather than respecting a specific job's configuration.

I've created an isolated case that displays the broken behavior and proves that LOC is the perpetrator:
```ruby
require "sidekiq"
require "sidekiq-unique-jobs"

Sidekiq.configure_server do |config|
  config.redis = { url: "redis://localhost:6379/15" }

  config.client_middleware do |chain|
    chain.add SidekiqUniqueJobs::Middleware::Client
  end

  config.server_middleware do |chain|
    chain.add SidekiqUniqueJobs::Middleware::Server
  end

  SidekiqUniqueJobs::Server.configure(config)
end

Sidekiq.configure_client do |config|
  config.redis = { url: "redis://localhost:6379/15" }
  config.client_middleware do |chain|
    chain.add SidekiqUniqueJobs::Middleware::Client
  end
end

class MyJob
  include Sidekiq::Worker
  sidekiq_options lock: :until_and_while_executing, lock_prefix: :foo
end

Sidekiq::Queue.new.each(&:delete)

MyJob.perform_async

puts(SidekiqUniqueJobs::Digests.new.entries)

Sidekiq::Queue.new.each(&:delete)

module SidekiqUniqueJobs
  module Job
    def add_lock_prefix(item)
      item[LOCK_PREFIX] ||= SidekiqUniqueJobs.config.lock_prefix
    end
  end
end 

MyJob.perform_async

puts(SidekiqUniqueJobs::Digests.new.entries)

Sidekiq::Queue.new.each(&:delete)
```

The output shows the initial lock contains the wrong prefix, while after the monkey-patch it uses the correct configuration:
![Screenshot 2024-07-18 154315](https://github.com/user-attachments/assets/3456c19e-e024-4d0c-97af-2b6510a1d2e7)

**Expected behavior**

The created lock should use the prefix set via the `lock_prefix` configuration option.

**Current behavior**

The created lock always uses the global default prefix.

**Additional context**
[...]

PR #886, opened 2025-08-01
Fix ignored custom lock prefix

luis30n:
This PR aims to fix the following issue: https://github.com/mhenrixon/sidekiq-unique-jobs/issues/850

Lock prefix argument is not being respected, hence all locks start by `uniquejobs` which makes it harder to debug or check the locks status of a given job as there's no way to identify them at first sight


mhenrixon (approved):
That looks correct, but I'll have to dig into this. 

It would be weird if no tests fail because of this change.

luis30n:
> That looks correct, but I'll have to dig into this.
> 
> It would be weird if no tests fail because of this change.

Thanks for reviewing this so quickly! 
I see coverage and rubocop actions failing...but errors seem unrelated to my changes.
I've added a spec to cover the custom lock_prefix logic, so that we'd detect future regressions on the CI

luis30n:
@mhenrixon would this cause issues with already acquired locks? I mean:
1. a lock is acquired as `uniquejobs:1234` on a rails app
2. we merge this, so lock_prefix starts to work as expected
3. such rails up is upgraded to latest gem version
4. another job with same lock args, which should be discarded, is enqueued, but the custom lock prefix is now taken into account so gem searches for `custom_prexi:1234`, does not find it and allows the job to be enqueued
5. both jobs run: unintentional duplicates

would this be an issue we should consider?

luis30n:
> @mhenrixon would this cause issues with already acquired locks? I mean:
> 
> 1. a lock is acquired as `uniquejobs:1234` on a rails app
> 2. we merge this, so lock_prefix starts to work as expected
> 3. such rails up is upgraded to latest gem version
> 4. another job with same lock args, which should be discarded, is enqueued, but the custom lock prefix is now taken into account so gem searches for `custom_prexi:1234`, does not find it and allows the job to be enqueued
> 5. both jobs run: unintentional duplicates
> 
> would this be an issue we should consider?

I understand that in other to migrate to using lock_prefix as expected without causing duplicates we should keep checking for the uniquejobs prefix as well on `.taken?` with something like:
```
def taken?(conn)
  conn.pipelined do |pipeline|
    pipeline.hexists(key.locked, job_id)
    pipeline.hexists(key.locked_fallback, job_id)
  end.any? { |exists| exists == 1 }
end
```
being locked fallback equivalent to lock but with "uniquejobs" prefix. WDYT?

luis30n:
@mhenrixon any news on this?

luis30n:
Ping @mhenrixon

PR #889, opened 2025-08-29
Correctly calculates expiration times in lua scripts.

brunoreisps:


texpert:
@mhenrixon, please, cherry-pick this fix into 8.x!

PR #892, opened 2025-10-30
Add support for lock_ttl to be a ActiveSupport::Duration

rlisowski:
`ActiveSupport::Duration` used to work. It stopped on [v8.0.11](https://github.com/mhenrixon/sidekiq-unique-jobs/tree/v8.0.11) when support for Proc/class method was added. Can we bring it back?

### DIFF
--- a/lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
+++ b/lib/sidekiq_unique_jobs/orphans/ruby_reaper.rb
@@ -331,8 +331,7 @@ module SidekiqUniqueJobs
       # @return [false] when missing
       #
       def in_sorted_set?(key, digest)
-        # Increased count from 1 to 50 for better throughput
-        conn.zscan(key, match: "*#{digest}*", count: 50).to_a.any?
+        conn.zscan(key, match: "*#{digest}*", count: 1).to_a.any?
       end
     end
     # rubocop:enable Metrics/ClassLength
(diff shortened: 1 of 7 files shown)


lib/sidekiq_unique_jobs/lock_ttl.rb
```diff
@@ -87,7 +87,7 @@ def fetch_ttl
 
     def calculate_timing(ttl)
       case ttl
-      when String, Numeric
+      when String, Numeric, ActiveSupport::Duration
```
mhenrixon: Wouldn't this crash if someone runs without active support?

rlisowski: Good point. Sorry for that. I added a check if `ActiveSupport::Duration` is defined.

mhenrixon:
I think we are going to have to do without adding the gem as a dependency. Just not sure how to test it. Maybe just define the constant in a test and let it behave as expected. What do you think?

danielmbarlow:
In case it helps someone I found that v8.0.11 worked and when upgrading to v8.0.12 the error appears. For the short term I added this to my project Gemfile:

```ruby
gem "sidekiq-unique-jobs", "~> 8.0.0", "< 8.0.12"
```

Thanks for working on this issue mhenrixon and rlisowski

danielmbarlow:
It seems that CI hasn't run because [this workflow](https://github.com/mhenrixon/sidekiq-unique-jobs/actions/runs/22917444367) is only waiting for another approval on this PR after the merge from `main`

image

